PR #795, opened 2014-12-09
Force backgroundRun in search

Warry:
screen shot 2014-12-09 at 16 50 30

### DIFF
--- a/ui/app/assets/services/sbt/tasks.js
+++ b/ui/app/assets/services/sbt/tasks.js
@@ -389,11 +389,11 @@ define([
           var redirected;
           // Hard-coded replacement of run -> backgroundRun
           if (command === "run") {
-            redirected = "run";
-            command = "backgroundRun";
+            redirected = runCommand();
+            command = "run";
           } else if (command === "runMain") {
-            redirected = "runMain";
-            command = "backgroundRunMain";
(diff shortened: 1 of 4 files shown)


ui/app/assets/services/sbt/tasks.js
```diff
@@ -384,18 +384,36 @@ define([
...
-            window.location.hash = "#build";
+
+          var command = message.partialCommand + completion.append;
+          var redirected;
+          // Hard-coded replacement of run -> backgroundRun
+          if (command === "run") {
+            redirected = runCommand();
+            command = "run";
+          } else if (command === "runMain") {
+            redirected = "bacgroundRunMain";
```
henrikengstrom: s/bacgroundRunMain/backgroundRunMain

henrikengstrom: Also, we don't want to use ``backgroundRunMain`` if it is a Play application. 
See logic here: https://github.com/typesafehub/activator/blob/wip/sbt-server/ui/app/assets/services/sbt/tasks.js#L133-146

Warry: if sbt autocomplete it, I have to show it. By typing 'run', it will run correctly

PR #820, opened 2014-12-22
Add a button to open a play app when the port is available

Warry:
screen shot 2014-12-22 at 15 34 54

### DIFF
--- a/ui/app/assets/services/sbt/tasks.js
+++ b/ui/app/assets/services/sbt/tasks.js
@@ -227,7 +227,7 @@ define([
       debug && console.log("TestEvent: ", event);
       execution.testResults.push(event.serialized);
     } else if (event.name === "PlayServerStarted") {
-      playApplicationUrl(event.serialized && event.serialized.url);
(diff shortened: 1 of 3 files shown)


ui/app/assets/services/sbt/tasks.js
```diff
@@ -123,6 +123,8 @@ define([
     return false;
   }
 
+  var playApplicationUrl = ko.observable();
```
havocp: init this to null or something?

ui/app/assets/services/sbt/tasks.js
```diff
@@ -224,6 +226,8 @@ define([
     } else if (event.name === "TestEvent") {
       debug && console.log("TestEvent: ", event);
       execution.testResults.push(event.serialized);
+    } else if (event.name === "PlayServerStarted") {
+      playApplicationUrl(event.serialized && event.serialized.url);
```
havocp: JS argh for returning a string and not a bool from this :-)

havocp:
this is probably fine to merge, though.

PR #822, opened 2015-01-06
Wip/h3nk3 echo integration

henrikengstrom:
Brings _echo_ and _sbt-echo_ into Activator.

Each project has its own build file and the projects have been put into respective folders (echo and sbt-echo).
They have been hooked into the overall build (_compile_ in **root** will compile the whole shabang).

There are still things to be done in terms of improvements, e.g. getting the version numbers aligned and in _one_ place, but this will be done in coming PRs. Also, cotests are failing atm - working on that.

Review by @corruptmemory or @havocp


havocp:
In build.scala, the "import org.apache.tools.ant.taskdefs.Echo" is probably just an IDE glitch?

havocp:
In build.scala, "import EchoBuild._" isn't really used I think

havocp:
We definitely need to clean up all the cut-and-paste versions in project/* but this LGTM as a start and then we can review the cleanups with a much smaller diff ;-)

henrikengstrom:
Thanks. Will clean up stuff wrt your comments above and also sort out the integration test issue. Once I am done with that I will merge and we can fix things in smaller PRs as you suggest.

corruptmemory:
Only other question I would have for this is: do we need to keep sigar?

henrikengstrom:
@corruptmemory if we don't need it anymore we should remove it, but that could be done in a future commit. Let's sit down and discuss what direction we want to take the Inspect feature. Sounds good?

PR #825, opened 2015-01-07
Upgrade to SBT 0.13.6

benmccann:


havocp:
Currently master is "commits that would make sense on both the activator-1.2.x and wip/sbt-server branches", where wip/sbt-server will become the new release series shortly (probably 1.3.x or 1.4.x). I think we've already bumped sbt and play, or shortly will, on wip/sbt-server, and we probably won't on 1.2.x (because it involves bumping them in a bunch of dependencies also, or else offlineTests will fail, so it's a bunch of work).

benmccann:
Okay, thanks for the update. Sounds like 1.3.x is close to being released, so feel free to close this then. I'm looking forward to projects created by activator using SBT 0.13.6 since previous versions have a bug that make tests do strange things which we get reports of ever so often with Play

PR #829, opened 2015-01-09
Moves Activator to the latest sbt version

henrikengstrom:
We need 0.13.7 of sbt as this is required by TRP.

Also adds load failure detection of project. 
This could need some UI love by @Warry.
Right now there is a pop-up fired up when a build failure is detected. A better variant would probably be to use a modal window with which we can do more things like reload into the home start page when the user clicks a button, link to the Typesafe subscription page, explain in detail how to set things up.

This is _not_ a generic solution for when the project fails to load as it is only warning about TRP projects loaded. We can always add information to the ``BuildFailedToLoad`` event with why the project did not load. This is a starting point that at least does _something_ to explain what is going on.

### DIFF
--- a/ui/app/assets/plugins/monitoring/newrelic/newrelic.html
+++ b/ui/app/assets/plugins/monitoring/newrelic/newrelic.html
@@ -61,9 +61,9 @@
 
     </div>
     <div class="tabs-step configure" data-bind="css: { opened: selectedTab() == 'configure' }">
-      <h2>Enter Licence Key</h2>
-      <p>To get your licence key, you must <a href="https://www.newrelic.com/typesafe" target="_blank">Sign Up</a> or <a href="https://rpm.newrelic.com/" target="_blank">Login</a> to a New Relic account.</p>
-      <p>Thereafter your licence key is available in the application setup screen (<a href="/public/plugins/monitoring/newrelic/LicenseKey.png" target="_blank">see this screenshot</a>)</p>
+      <h2>Enter License Key</h2>
+      <p>To get your license key, you must <a href="https://www.newrelic.com/typesafe" target="_blank">Sign Up</a> or <a href="https://rpm.newrelic.com/" target="_blank">Login</a> to a New Relic account.</p>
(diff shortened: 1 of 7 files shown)


ui/app/assets/plugins/monitoring/newrelic/newrelic.html
```diff
@@ -61,9 +61,9 @@
 
     </div>
     <div class="tabs-step configure" data-bind="css: { opened: selectedTab() == 'configure' }">
-      <h2>Enter Licence Key</h2>
-      <p>To get your licence key, you must <a href="https://www.newrelic.com/typesafe" target="_blank">Sign Up</a> or <a href="https://rpm.newrelic.com/" target="_blank">Login</a> to a New Relic account.</p>
-      <p>Thereafter your licence key is available in the application setup screen (<a href="/public/plugins/monitoring/newrelic/LicenseKey.png" target="_blank">see this screenshot</a>)</p>
+      <h2>Enter License Key</h2>
```
corruptmemory: These fixes should be on a separate PR.  j/k - ignore :-)

ui/app/assets/widgets/home/home.js
```diff
@@ -2,29 +2,39 @@
...
+    window.alert('Could not load project.\n' +
+    'If you are creating Typesafe Reactive Platform project you must add a \'typesafe.propertied\' file in the \'<template>/project\' folder.\n' +
+    'The file must contain your subscription id in the following format \'typesafe.subscription=<YOUR ID>\'\n' +
+    'For more information see: http://typesafe.com/subscription');
+  });
+
   var SharedState = {
     working: ko.observable(false)
-  }
+  };
```
corruptmemory: Thank you IntelliJ

corruptmemory:
LGTM

ui/app/assets/plugins/monitoring/newrelic/newrelic.html
henrikengstrom: Hey, at least they are in a separate commit! ;)

PR #833, opened 2015-01-12
Do not show Inspect as activated until sbt has restarted and confirmed

Warry:
screen shot 2015-01-12 at 18 14 52


henrikengstrom:
This does not work as intended.

When I try to enable Inspect on "hello-akka" I see the pop-up when we try to install the plugin but that is immediately replaced with a pop-up saying "Inspect is not available".

henrikengstrom:
@Warry I cannot seem to trigger the popup anymore. What did you do to test it?

Warry:
@henrikengstrom TL;DR: it is working, you have to click run. 

Here is the thing: To determine if the project supports inspect or not, I'm waiting for the "whyInspectIsNotSupported" value (based on version events).

Previously, after the activation of Inspect I started tracking the end of the next/current build session and lookup if the  "whyInspectIsNotSupported" value was set in the mean time. That way I could know after installing Inspect if it was supported.

However, sometimes I had sequential builds where the first one didn't raise any version. So the value didn't change... So as seen with Havoc, I'm now doing the checking on clicking on run.

henrikengstrom:
Well the issue is that I couldn't get the popup to show. What project did you use to trigger it? I would like to test that it seems ok. Code wise everything looks fine.

Warry:
@henrikengstrom If you run hello-scala then activate Inspect, you can see the popup by clicking Run

henrikengstrom:
LGTM

PR #847, opened 2015-01-20
Ask for typesafe id in a popup before cloning

Warry:
if the chosen template is TRP, then it asks for an ID then send it with the `CreateNewApplication` message (the associate json object has a `typesafeId` field).

### DIFF
--- a/ui/app/assets/widgets/home/templates/templates.js
+++ b/ui/app/assets/widgets/home/templates/templates.js
@@ -6,18 +6,49 @@ define([
   'commons/websocket',
   'widgets/fileselection/fileselection',
   'text!./templates.html',
+  'widgets/modals/modals',
   'css!./templates'
 ], function(
   settings,
   websocket,
   FileSelection,
-  tpl
+  tpl,
+  modals
 ) {
 
   var trpInfoSeen = settings.observable("reactive-platform.accepted-license", false);
 
+  var typesafeId = settings.observable("TypesafeID", "");
+  var activeOk = ko.observable(!!typesafeId());
+  function saveTypesafeID(e){
+    if (e.target.value.length === 36){
+      activeOk(true);
+      typesafeId(e.target.value);
+    } else {
+      activeOk(false);
+      typesafeId("");
+    }
+  }
+  function askForTypesafeId(callback){
+    var message = $("<article/>").html("<p>You are creating Typesafe Reactive Platform project, which require a Typesafe ID.</p><p>You can retreive you ID on the <a href='http://typesafe.com/subscription' target='_blanc'>typesafe.com/subscription</a> page.<p><p class='input'></p>")[0];
+    $("<input class='typesafeId-form' type='text' />").change(saveTypesafeID).keyup(saveTypesafeID).val(typesafeId()).appendTo($(".input", message));
+    modals.show({
+      shape: "large",
+      title: "Submit your Typesafe ID",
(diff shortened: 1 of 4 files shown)


ui/app/assets/widgets/home/templates/templates.js
```diff
@@ -6,19 +6,50 @@ define([
...
+    if (e.target.value.length === 36){
+      activeOk(true);
+      typesafeId(e.target.value);
+    } else {
+      activeOk(false);
+      typesafeId("");
+    }
+  }
+  function askForTypesafeId(callback){
+    var message = $("<article/>").html("<p>You are creating Typesafe Reactive Platform project, which require a Typesafe ID.</p><p>You can retreive you ID on the <a href='http://typesafe.com/subscription' target='_blanc'>typesafe.com/subscription</a> page.<p><p class='input'></p>")[0];
```
havocp: s/require/requires/ s/retreive/retrieve/ s/you/your/

We should suggest the same page as the RP plugin's error in the logs; does it suggest /subscription or /account/id ?

havocp: Let's say "retrieve your ID or sign up for a free trial" ?

ui/app/assets/widgets/home/templates/templates.js
```diff
@@ -6,19 +6,50 @@ define([
...
+  tpl,
+  modals
 ) {
 
   var trpInfoSeen = settings.observable("reactive-platform.accepted-license", false);
 
+  var typesafeId = settings.observable("TypesafeID", "");
+  var activeOk = ko.observable(!!typesafeId());
+  function saveTypesafeID(e){
+    if (e.target.value.length === 36){
```
havocp: trimming whitespace off the string for people might save some confusion for them

havocp:
I have the card to implement the backend for this, which should be simple enough - will do as soon as this is merged.

ui/app/assets/widgets/home/templates/templates.js
```diff
@@ -6,19 +6,50 @@ define([
...
 ], function(
   settings,
   websocket,
   FileSelection,
-  tpl
+  tpl,
+  modals
 ) {
 
   var trpInfoSeen = settings.observable("reactive-platform.accepted-license", false);
```
havocp: btw is this now obsolete? I guess our current idea is that if you have a subscription ID, that means you accepted the license (you have to accept in order to get an ID)

ui/app/assets/widgets/home/templates/templates.js
```diff
@@ -6,19 +6,50 @@ define([
...
 
+  var typesafeId = settings.observable("TypesafeID", "");
+  var activeOk = ko.observable(!!typesafeId());
+  function saveTypesafeID(e){
+    if (e.target.value.trim().length === 36){
+      activeOk(true);
+      typesafeId(e.target.value);
```
havocp: you trim before validating but then haven't trimmed what you store here, right?

ui/app/assets/widgets/home/templates/templates.js
Warry: We still show the pop up on the very first visit of that tab !

havocp: ah, ok. let's leave it alone for now, not mix with this PR

PR #851, opened 2015-01-21
Divides integration tests in Echo apart from the normal tests and runs t...

henrikengstrom:
...hem separately.

**Please do not merge just yet!**


henrikengstrom:
Closing this. Will make a new attempt once a version of sbt-rc with pickling is available.

PR #852, opened 2015-01-22
Wip/h3nk3 new template cache

henrikengstrom:


project/build.scala
```diff
@@ -108,7 +108,6 @@ object TheActivatorBuild extends Build {
     settings(PlayKeys.playDefaultPort := 8888)
     settings(Keys.includeFilter in (Assets, LessKeys.less) := "*.less")
     settings(Keys.excludeFilter in (Assets, LessKeys.less) := "_*.less")
-    settings(Keys.initialize ~= { _ => sys.props("scalac.patmat.analysisBudget") = "512" })
```
corruptmemory: Why is this no-longer needed?  What caused this to be needed before?

henrikengstrom: Seems you added this waaaay back: https://github.com/typesafehub/activator/commit/ce007c6eca60535c334e0aea37e73cc71d27adf5

The reason for not using this anymore is we get tons of this warning in the build right now:
``
[warn] /Users/he/code/typesafe/activator/echo/trace/event/src/main/scala/com/typesafe/trace/util/ProtobufConverter.scala:1187: Please remove -Dscalac.patmat.analysisBudget, it is ignored.
[warn]   private[util] def fromProtoSystemMessage(proto: ProtoAnnotationData): SysMsg = proto.getSysMsg match {
[warn]                     ^
[warn] /Users/he/code/typesafe/activator/echo/trace/event/src/main/scala/com/typesafe/trace/util/ProtobufConverter.scala:1204: Please remove -Dscalac.patmat.analysisBudget, it is ignored.
[warn]   private[util] def fromProtoInfo(info: ProtoInfo): Info = info.getType match {
``

corruptmemory: Interesting, something must have changed in scala.

corruptmemory:
LGTM

project/build.scala
gkossakowski: See scala/scala#4078

henrikengstrom: Thanks for the explanation @gkossakowski

PR #855, opened 2015-01-23
Mark opened files as dirty if parent folder or file itself was renamed

Warry:
I cannot easily track the change on the particular file, because it would require to change a lot of static / immutable / working things to become async and observable (such as: type of the file, syntax highlighting...). So just showing it as dirty, like mot text editors.


henrikengstrom:
Since this is not an IDE I think the "dirty feature" is good enough.
LGTM

PR #856, opened 2015-01-23
Re-wire IDE files generation

Warry:


ui/app/assets/plugins/code/code.html
```diff
@@ -5,6 +5,10 @@
   <div class="hint code">
     <p>Browse your app's code</p>
     <span>You can also open it in your favorite editor the same way</span>
+    <p>
+      <button data-bind="click: openInEclipse, disbale: eclipseReady">Open in Eclipse</button>
+      <button data-bind="click: openInIdea, disbale: ideaReady">Open in Idea</button>
```
henrikengstrom: To clarify I think we should use ``IntelliJ`` instead of ``Idea`` here.

henrikengstrom: s/disbale/disable

ui/app/assets/plugins/code/eclipseGenerator.js
```diff
@@ -1,30 +1,57 @@
...
+  var isInstalled = ko.observable(false);
+  function checkInstalled() {
+    fs.exists(fs.absolute(projectFile), function(r) {
+      isInstalled(r.location);
+    });
+  }
+  checkInstalled();
+
   var generate = function(overrideExisting) {
+    if (isInstalled() && !window.confirm("Eclipse is already installed, do you want to re-grenerate the project?")) return;
```
henrikengstrom: s/re-grenerate/regenerate

ui/app/assets/plugins/code/generator.js
```diff
@@ -73,12 +75,14 @@ define([
...
         // Look for a ".project" file in the home directory to see if there already is an existing Eclipse project
-        ajax.browse(serverAppModel.location + "/" + generator.projectFile).done(function (data) {
-          resetState("Required file(s) exist. Open your IDE and import project.");
-        }).error(function () {
-          // No project files found - continue the process
-          logs.push({message: "No project file(s) found - continuing the process."});
-          checkCommand();
+        ajax.exists(serverAppModel.location + "/" + generator.projectFile, function (status) {
+          if (status === "success") {
+            resetState("Required file(s) exist. Open your IDE and import project.");
```
henrikengstrom: Perhaps a better way to put this is "Required file(s) created."?

henrikengstrom: Cancel that. It was for already existing files and then it makes sense to use "exist".

ui/app/assets/plugins/code/ideaGenerator.js
```diff
@@ -1,30 +1,57 @@
...
+  var isInstalled = ko.observable(false);
+  function checkInstalled() {
+    fs.exists(fs.absolute(projectFile), function(r) {
+      isInstalled(r.location);
+    });
+  }
+  checkInstalled();
+
   var generate = function(overrideExisting) {
+    if (isInstalled() && !window.confirm("Idea is already installed, do you want to re-grenerate the project?")) return;
```
henrikengstrom: ``IntelliJ`` instead of ``Idea``?
s/re-grenerate/regenerate

ui/app/assets/plugins/code/eclipseGenerator.js
henrikengstrom: I think it should say "Eclipse project already exists. Do you want to regenerate the project?"

ui/app/assets/plugins/code/ideaGenerator.js
henrikengstrom: I think it should say "IntelliJ project already exists. Do you want to regenerate the project?"

ui/app/assets/widgets/browser/browser.html
```diff
@@ -11,6 +11,8 @@
           <a data-bind="click: newFileAtRoot">New file</a>
           <a data-bind="click: refreshProject">Refresh project</a>
           <a data-bind="click: revealProject">Reveal in system</a>
+          <a data-bind="click: openInEclipse">Open in Eclipse</a>
+          <a data-bind="click: openInIdea">Open in Idea</a>
```
henrikengstrom: ``Idea`` => ``IntelliJ``

ui/app/assets/plugins/code/code.html
henrikengstrom: Perhaps better wording is "Create X project" where X = Eclipse || IntelliJ?
Open to mean sounds like somehow IntelliJ/Eclipse will be opened with the project loaded.

henrikengstrom:
Almost not worth adding the IntelliJ plugin as it generates a deprecated project layout for v14. The following warning is issued:
``The project 'hello-akka' has an older format and will be converted. You may not be able to open the project with earlier versions of IntelliJ IDEA. Details...  Old versions of project files will be saved to: '/Users/he/hello-akka/projectFilesBackup'``

I guess we cannot count on that everyone is using IntelliJ v14 though.

Warry:
What you mean @henrikengstrom ? That we should disable Idea for a while?

henrikengstrom:
@Warry let's keep it around and we can always remove it if/when people complain too much. It all depends on what version of IntelliJ they are using. < 14 it's all good since only v14 is the one that can auto import sbt projects.

henrikengstrom:
LGTM

PR #858, opened 2015-01-27
On startup, add activator offline repo to ~/.sbt/repositories

havocp:
This is so sbt server and command line sbt will find it.

cc @jsuereth

### DIFF
--- a/launcher/src/main/scala/activator/RepositoryConfig.scala
+++ b/launcher/src/main/scala/activator/RepositoryConfig.scala
@@ -1,0 +1,160 @@
+/**
+ * Copyright (C) 2015 Typesafe <http://typesafe.com/>
+ */
+package activator
+
+import scala.util.control.NonFatal
+import activator.properties.ActivatorProperties._
+import java.io.File
+
+object RepositoryConfig {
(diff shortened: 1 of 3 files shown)


launcher/src/main/scala/activator/RepositoryConfig.scala
```diff
@@ -0,0 +1,160 @@
...
+    val uriString = (new java.io.File(path)).toURI.toASCIIString()
+    if (uriString.startsWith("file://"))
+      uriString.substring("file://".length)
+    else
+      path // give up, hope for best?
+  }
+
+  private def newRepositorySection(oldOption: Option[Section]): Section = {
+    val activatorLocalLine =
+      """  activator-local: file://${activator.local.repository-""" +
```
henrikengstrom: Not sure I understand this layout? Is it intended to look like ``file://${activator.local.repository-/Users/x/y/z/repository``?

henrikengstrom: I guess it is: https://github.com/typesafehub/activator/blob/wip/offline-repo-config/project/integration.scala#L151

henrikengstrom:
LGTM

PR #860, opened 2015-01-27
Wip/h3nk3 shims

henrikengstrom:
Functionality that adds sbt Play plugin automatically to a project.
Also contains some upgrades of stale versions.

Will fix the fact that versions are hardcoded to a better way of handling them in separate PR.

### DIFF
--- a/ui/app/assets/services/sbt/configuration.js
+++ b/ui/app/assets/services/sbt/configuration.js
@@ -22,9 +22,14 @@ define([
   var echoPluginFileContent = "// This plugin runs apps with the \"echo\" trace infrastructure which backs up the Inspect functionality in Activator\n\n" +
     "addSbtPlugin(\"com.typesafe.sbt\" % \"sbt-echo\" % \"0.1.7\")";
 
+  var playForkRunPluginFileLocation = "/project/play-fork-run.sbt";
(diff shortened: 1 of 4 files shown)


ui/app/assets/services/sbt/configuration.js
```diff
@@ -78,11 +84,16 @@ define([
...
+  checkFileContent(serverAppModel.location+playForkRunPluginFileLocation, playForkRunPluginFileContent, function () {
+    addedPlayForkRun(true);
+  });
+
   return {
     echoInstalledAndReady: echoInstalledAndReady,
     addedEchoFile:         addedEchoFile,
     addedBackgroundFile:   addedBackgroundFile,
     addedForkInRun:        addedForkInRun,
+    addedPlayForkRun:      addedPlayForkRun,
```
havocp: it seems like there should be a change elsewhere to use this - wherever we are using the other similar observables?

havocp: it's the echoReady above in this same file where we use them - I don't actually know why any of these "added" things are exported from the module.

henrikengstrom: You're right. I just followed the pattern to "expose" it in case there is a future need to hook it in somewhere.

PR #863, opened 2015-01-28
update to identify TaskSuccess by $type

havocp:
There's no longer a boolean "success" field, which is sort
of ugly in JS but avoids a special case in Scala


henrikengstrom:
LGTM

PR #865, opened 2015-01-28
Fixed shim for sbt-echo

corruptmemory:


ui/app/assets/services/sbt/configuration.js
```diff
@@ -20,7 +20,8 @@ define([
 
   var echoPluginFileLocation = "/project/inspect.sbt";
   var echoPluginFileContent = "// This plugin runs apps with the \"echo\" trace infrastructure which backs up the Inspect functionality in Activator\n\n" +
-    "addSbtPlugin(\"com.typesafe.sbt\" % \"sbt-echo\" % \"0.1.7\")";
+    "addSbtPlugin(\"com.typesafe.play\" % \"sbt-fork-run-plugin\" % \"2.3.8-M1\")\n\n" +
```
henrikengstrom: How is this different from the lines below (26-28)?

ui/app/assets/services/sbt/configuration.js
```diff
@@ -20,7 +20,8 @@ define([
...
+    "addSbtPlugin(\"com.typesafe.sbt\" % \"sbt-echo-play\" % \"0.1.8\")";
```
henrikengstrom: This one I get though.

ui/app/assets/services/sbt/configuration.js
corruptmemory: Yeah, I went back looking at the code and how all the checks and wirings are made and I see I have this in the wrong place.  Basically, what we need is this:

If Play: emit this into some shim:

```scala
addSbtPlugin("com.typesafe.play" % "sbt-fork-run-plugin" % "2.3.8-M1")

addSbtPlugin("com.typesafe.sbt" % "sbt-echo-play" % "0.1.8")
```

else (not play) emit this:

```scala
addSbtPlugin("com.typesafe.sbt" % "sbt-echo" % "0.1.8")
```

corruptmemory:
Need to rethink how this is done (see my comment).  Closing.

henrikengstrom:
The way I understood it was that we always could add this plugin: 
``addSbtPlugin("com.typesafe.play" % "sbt-fork-run-plugin" % "2.3.8-M1")``

But I see the issue with ``sbt-echo`` vs ``sbt-echo-play``.
You have this convenience method to find out in JS if it is a Play app or not:
https://github.com/typesafehub/activator/blob/wip/pickling-based/ui/app/assets/services/sbt/tasks.js#L120-126

PR #869, opened 2015-01-29
fix ui regressions

Warry:


havocp:
Some of these changes predate the pickling conversion - I didn't realize we were so far behind. Anyway, lgtm.

PR #871, opened 2015-01-29
Improve logs ui performances

Warry:


ui/app/assets/widgets/home/working/working.js
```diff
@@ -9,40 +9,45 @@ define([
...
+
+  var State = {
+    logs: logs,
+    scroll: scroll
+  }
+
+  websocket.subscribe({type: "sbt", subType: "CoreLogEvent"}).fork().each(function(message) {
+    logs.push({
+      message: message.event.entry.message,
+      type: "info"
```
havocp: is this deliberate to always use info log level?

havocp:
This is a nice focused PR, it looks clean.

ui/app/assets/widgets/home/working/working.js
Warry: Yeah, there is no specific styling in those logs anyway.

ui/app/assets/plugins/run/system/system.html
```diff
@@ -18,10 +18,11 @@
         </div>
       </div>
       <div class="logs" data-bind="logScroll: memoLogsScroll">
-        <ul data-bind="foreach: sbt.logs.stdout">
+        <ul data-bind="logEach: sbt.logs.stdout">
           <li data-bind="text: event.entry.message, attr: { 'data-level': event.entry.level, 'data-type': event.entry.type }"></li>
```
havocp: does event.entry.type need to be event.entry.$type

ui/app/assets/widgets/home/working/working.js
havocp: event.entry.level certainly can exist in CoreLogEvent, whether you happened to get a non-info one or not...

ui/app/assets/plugins/run/system/system.html
Warry: that's one of the regression I was mentioning, earlier in the call.

havocp: I think we're going to stick to $type so the whole protocol is consistent, even though it is uglier in JS. so we should go ahead and update this.

Warry: Oh sorry. gotcha !

PR #872, opened 2015-01-29
Fixed inspect shims

corruptmemory:
* Also got running with echo working

Sadly, inspect actually not working at the moment.  Still need to debug
this :-(

### DIFF
--- a/ui/app/snap/SbtClientActor.scala
+++ b/ui/app/snap/SbtClientActor.scala
@@ -60,11 +60,6 @@ class SbtClientActor(val client: SbtClient) extends Actor with ActorLogging {
     }
     case structure: MinimalBuildStructure =>
       forwardOverSocket(BuildStructureChanged(structure))
-    case PlayAvailable(available) =>
-      context.parent ! NotifyWebSocket(SbtProtocol.wrapEvent(
-        JsObject(Seq(
-          "backgroundRunnerAvailable" -> JsBoolean(available))),
-        "PlayStatus"))
     case req: ClientAppRequest => {
       req match {
         case re: RequestExecution =>
@@ -115,7 +110,6 @@ class SbtClientLifeCycleHandlerActor(val client: SbtClient) extends Actor with A
       handleEvents
       watchBuild
       setupSubscription
-      checkPlayAvailable
   }
 
   override def postStop(): Unit = {
@@ -138,14 +132,6 @@ class SbtClientLifeCycleHandlerActor(val client: SbtClient) extends Actor with A
     buildSub = Some(client.watchBuild { structure =>
       context.parent ! structure
     })
-  }
-
-  def checkPlayAvailable = {
-    client.lookupScopedKey("backgroundRun") map { keys: Seq[ScopedKey] =>
-      if (keys.length > 0) {
-        context.parent ! SbtClientActor.PlayAvailable(true)
-      }
-    }
   }
 
   // this is a hardcoded hack... we need to control the list of things
--- a/ui/app/snap/SbtClientActor.scala
+++ b/ui/app/snap/SbtClientActor.scala
@@ -141,7 +141,7 @@ class SbtClientLifeCycleHandlerActor(val client: SbtClient) extends Actor with A
   }
 
   def checkPlayAvailable = {
-    client.lookupScopedKey("playBackgroundRunTaskBuilder") map { keys: Seq[ScopedKey] =>
+    client.lookupScopedKey("backgroundRun") map { keys: Seq[ScopedKey] =>
       if (keys.length > 0) {
         context.parent ! SbtClientActor.PlayAvailable(true)
       }
--- a/sbt-echo/play/src/main/scala/com/typesafe/sbt/SbtEchoPlay.scala
+++ b/sbt-echo/play/src/main/scala/com/typesafe/sbt/SbtEchoPlay.scala
@@ -25,7 +25,8 @@ object SbtEchoPlay extends AutoPlugin with PlayInternalKeys {
   override def requires = Play && PlayForkRun && SbtEcho
 
   def playTraceJavaOptionsTask: Def.Initialize[Task[Seq[String]]] = Def.task {
-    echo.EchoRun.traceJavaOptions(aspectjWeaver.value, sigarLibs.value)
+    echo.EchoRun.traceJavaOptions(aspectjWeaver.value, sigarLibs.value) ++
+      Seq("-Dconfig.resource=application.conf")
   }
 
   def playForkOptionsTask: Def.Initialize[Task[PlayForkOptions]] = Def.task {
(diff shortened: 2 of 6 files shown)


ui/app/snap/SbtClientActor.scala
```diff
@@ -141,7 +141,7 @@ class SbtClientLifeCycleHandlerActor(val client: SbtClient) extends Actor with A
   }
 
   def checkPlayAvailable = {
-    client.lookupScopedKey("playBackgroundRunTaskBuilder") map { keys: Seq[ScopedKey] =>
+    client.lookupScopedKey("backgroundRun") map { keys: Seq[ScopedKey] =>
```
henrikengstrom: So if ``backgroundRun`` is available this means that Play is available?

havocp: backgroundRun should be in all projects.

the technically-correct way to check for play is to look for the play plugin when we get BuildStructureChanged event, I think. 
https://github.com/sbt/sbt-remote-control/blob/master/commons/protocol/src/test/resource/saved-protocol/0.1/event/build_structure_changed.json  
Since it's per-project, for this purpose we probably want "use the play-enabled echo if any projects are play-based" and then the play-related echo should probably be made to work on akka-only projects if it doesn't.

corruptmemory: Actually, this is to simply determine if you have the right version of Play.  Which is more of a problem now since the backgroundRun is not part of play itself but part of the play forked dev-runner plugin.  So, we need some other check here because the fork dev runner is an auto plugin.  My suggestion will be to simply add a flag task to Play's plugin, something like 'supportsBackgroundRun' or so that we can test for.

Thoughts?

henrikengstrom: I thought we would not support any Play projects that do not have backgroundRun available? What do you mean by "right version" in this context?

corruptmemory: Our shim will add the sbt-fork-run-plugin to any project we deem a "Play project".  Since it's an autoplugin it will happily try to "attach itself" to any an all Play sub projects in your build.  This will have the effect of making versions of Play less than 2.3.8 seem like they have background run support, which they won't.  In fact, the build will probably barf in that case.  We can also use the results of the `echo:echoPlayVersionReport` but that implies that we add the echo (inspect) shim to get this report before deciding if we can actually determine if you have a version of Play that can be run in the background.

havocp: I wonder if our shim will actually force the play plugin to upgrade. i.e. you add the play-fork-run plugin 2.3.8, it will depend on play plugin 2.3.8, and ivy may evict play plugin < 2.3.8. If this is what happens, it's probably what we want...

corruptmemory: As of now, the way it works is we detect if a version of Play is compatible if it can be run in the background.  The presence of `playBackgroundRunTaskBuilder` used to be a reliable test because it was directly bundled with Play's sbt plugin.  Now the equivalent functionality is supplied by an auto plugin that will happily attach itself to any Play sub project and magic up a background run that won't work because the requisite changes to how the `BuildLink` and reloader for forked Play instances will be missing.  Make sense?

havocp: what I'm hoping for is that if the project has ANY version of play plugin, we could add our shim, which would force-upgrade the play plugin to 2.3.8. Then we don't have to detect whether the version of play is compatible.

henrikengstrom: @corruptmemory makes sense. Just wanted to know what the deal was. Thanks for the explanation. 
However, by using ``backgroundRun`` as a mean of checking Play means that all projects will be considered Play apps doesn't it?

corruptmemory: I just tested this by setting the Play plugin to 2.3.4 in my project.  The shims get added.  The first time the project is loaded I see:

```
Resolving com.typesafe.play#sbt-plugin;2.3.4 ...
Resolving com.typesafe.play#sbt-plugin;2.3.4 ...
Resolving com.typesafe.play#build-link;2.3.4 ...
Resolving com.typesafe.play#build-link;2.3.4 ...
Resolving com.typesafe.play#play-exceptions;2.3.4 ...
Resolving com.typesafe.play#play-exceptions;2.3.4 ...
Resolving org.javassist#javassist;3.18.2-GA ...
Resolving com.typesafe.play#routes-compiler_2.10;2.3.4 ...
Resolving com.typesafe.play#routes-compiler_2.10;2.3.4 ...
```

After the reload with the shims I see:

```
Read from stdout: Resolving com.typesafe.play#play_2.10;2.3.8-M1 ...
Resolving com.typesafe.play#play_2.10;2.3.8-M1 ...
Read from stdout: Resolving com.typesafe.play#build-link;2.3.8-M1 ...
Resolving com.typesafe.play#build-link;2.3.8-M1 ...
Read from stdout: Resolving com.typesafe.play#play-exceptions;2.3.8-M1 ...
Resolving com.typesafe.play#play-exceptions;2.3.8-M1 ...
Read from stdout: Resolving org.javassist#javassist;3.18.2-GA ...
Resolving org.javassist#javassist;3.18.2-GA ...
Read from stdout: Resolving com.typesafe.play#play-iteratees_2.10;2.3.8-M1 ...
```

Which does seem to indicate that the "automagical upgrade" is occurring.

henrikengstrom: @havocp that would be ideal

havocp: I think the automagical upgrade is a pretty good outcome, especially if it warns that 2.3.4 was evicted, but even if not. Certainly simpler to understand than the alternatives.

corruptmemory: Also the play plugin gets updated:

```
Resolving com.typesafe.play#sbt-plugin;2.3.8-M1 ...
```

corruptmemory: I don't see any eviction warning.

henrikengstrom: I don't get any eviction warnings either but it does use the Play version we specify in ``play-fork-run.sbt``.

havocp: Looking at the JavaScript that uses this PlayStatus event, I'm pretty sure we can delete this and also the usage of it in task.js. The usage is here:
https://github.com/typesafehub/activator/blob/wip/sbt-server/ui/app/assets/services/sbt/tasks.js#L539
and https://github.com/typesafehub/activator/blob/wip/sbt-server/ui/app/assets/services/sbt/tasks.js#L520

but play actually sets a main class: https://github.com/typesafehub/activator/blob/wip/sbt-server/ui/app/assets/services/sbt/tasks.js#L121

I think we could delete all this - where we use playHasRunCommand, just don't, and go by mainClass?

If we need to know whether we have a play app, then the "main class is netty server" check is a fine hack, and looking at BuildStructureChanged event for whether the play plugin is on the project is less hacky, but I don't know why we need this special server-side look at a key.

corruptmemory: Consider it deleted.

henrikengstrom:
LGTM, but...
Before we merge this one has to go into wip/sbt-server: https://github.com/typesafehub/activator/pull/875
Also, we publish v0.1.8 of Echo and sbt-echo before merging as things will break otherwise.

henrikengstrom:
And now echo and sbt-echo have been published so merging this one.

PR #873, opened 2015-01-29
Requests monitor in inspect

henrikengstrom:
Brings back the UI part for Play monitoring. The code has not been tested with a real Inspect integration yet as Play monitoring is not yet available. This lays a foundation for the UI and it might have to be improved once the integration is done.

Review by @Warry to verify that the layout is still valid (since the original commit was done over a month ago).


henrikengstrom:
LGTM

PR #882, opened 2015-02-02
Upgrades to latest version of mimeutil

henrikengstrom:
Haven't seen the issue after this upgrade, but it didn't happen every time before either so it might just be plain luck. Can't hurt to upgrade regardless.


corruptmemory:
LGTM

PR #884, opened 2015-02-03
Bumped versions

corruptmemory:
* Play -> 2.3.8-M2
* Echo -> 0.1.9 (published)
* Sbt-Echo -> 0.1.9 (Published)

Also fixed immediate compilation build problems.  Activator UI-prelated
problems may linger


havocp:
LGTM but it will break the JavaScript. maybe we should fix that in this same PR to avoid hosing master.
The changes I remember are `event.name` to `event.serialized.$type` (and add package scope), and for tests, `.error` becomes `.error.message`

Fortunately the compiler will just find these cases. just kidding.

havocp:
Merged with latest master and JS fixed up, look good to you?

PR #886, opened 2015-02-04
Exports versions to JS file for consumption by other JS files.

henrikengstrom:


project/VersionGenerator.scala
```diff
@@ -0,0 +1,39 @@
...
+         |    playVersion: playVersion,
+         |    echoVersion: echoVersion,
+         |    ideaVersion: ideaVersion,
+         |    eclipseVersion: eclipseVersion,
+         |    sbtUIVersion: sbtUIVersion
+         |  };
+         |});""".stripMargin
+
+    println(s"Generating file: ${path}${fileLocation}")
+    sbt.IO.write(file = new File(path + fileLocation), content = content)
```
havocp: in Properties.scala there's a writeIfChanged that would be good to use here perhaps.

havocp:
looks good other than the one comment. Are there any hardcoded versions in scala or are they all in JS?

henrikengstrom:
The only backend "hardcoded" versions I can think of are for AD/NR and they're defined in ``application.conf``

PR #888, opened 2015-02-04
Prevents publishing echo and sbt-echo from root project

henrikengstrom:


havocp:
this removes them from the offline repo but does it also block them from publishing via maven?

henrikengstrom:
I tried publishing locally and the change above prevented ``echo`` and ``sbt-echo`` from being published.

havocp:
I think it might work because it removes echo and sbt-echo from the root aggregation. The problem with that is that "test" and "compile" won't include echo or sbt-echo either, I think.

We can disable aggregation for publish (described here http://www.scala-sbt.org/0.13/tutorial/Multi-Project.html ) but that doesn't do what we want since then publish won't publish any subprojects.

We want to disable aggregation of publish/publishLocal/publishSigned/publishSignedLocal for echo, while still aggregating test and compile, and still allowing publish from within the subproject. Maybe `aggregate in publish in sbtEcho := false` would work but seems like a long shot.

Another idea would be to set a separate version number for echo and simply publish everything at the same time (make the publish aggregation work, instead of disabling it). Or use the Activator version number for echo, even? If we have a separate version, we have to be sure we automatically enforce that it's always incremented when we publish so we don't overwrite already-published jars.

henrikengstrom:
Ah, so this turned out to become more complicated than I had hoped. Did not verify that ``echo`` and ``sbt-echo`` were still part of the overall build.

Having all projects use the same version numbering is fine, but it means that there will be new releases of ``echo`` and ``sbt-echo`` every time we ship a new Activator. This may be fine though?

havocp:
Thinking about this overnight, I think removing echo from aggregation entirely might be ok for now (the main problem is it won't get Travis'd). Or, I'm also fine with republishing echo whenever we publish activator; I don't see an issue with that (?). We could also try `aggregate in publish in sbtEcho := false` on the off chance it works.

PR #890, opened 2015-02-05
Dependencies that make activator-offline-test pass the offline tests

henrikengstrom:
These changes make the ``activator-offline-test`` template work when testing offline. If the dependencies here below are enough to cover the featured templates it will mean an offline distribution of about 390MB.

To test:
``> sbt``
``> set LocalTemplateRepo.overrideWithTemplates in TheActivatorBuild.localTemplateRepo := "activator-offline-test"``
``> offlineTests```

It is important to override the default featured templates as none of them have been updated to contain the required versions of Activator.


havocp:
It's a good starting point - I suspect we'll have to put some of it back when we have the full template set, but we'll cross that bridge upon arrival.

PR #891, opened 2015-02-05
Version alignment

henrikengstrom:
This removes the two sbt version warnings shown when starting Activator.


corruptmemory:
LGTM

PR #892, opened 2015-02-05
Prevents the 'No dynamic serializer found' warning from being shown in t...

henrikengstrom:
...he build view


havocp:
did you grep for libraryDependencies in JS files? lgtm as long as we check for usage in JS.

havocp:
grepped, looks good

henrikengstrom:
Yes, I couldn't find any usage either.

PR #894, opened 2015-02-06
Improve logs performances when coming back and forth

Warry:
- Filters out debug, when not activated
- Caches the dom

### DIFF
--- a/ui/app/assets/services/sbt/logs.js
+++ b/ui/app/assets/services/sbt/logs.js
@@ -29,15 +29,24 @@ define([
     }
   }
 
+  function filterDebug(m) {
+    if (m.event.entry && m.event.entry.level)
+      return m.event.entry.level !== "debug" || (app.settings.showLogDebug() || debug);
+    else
+      return true;
+  }
+
   /**
   Logs, by execution/task
   */
   logEvent.matchOnAttribute("subType", "TaskLogEvent")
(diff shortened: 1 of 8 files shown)


ui/app/assets/services/sbt/logs.js
```diff
@@ -29,15 +29,24 @@ define([
...
+      return m.event.entry.level !== "debug" || (app.settings.showLogDebug() || debug);
+    else
+      return true;
+  }
+
   /**
   Logs, by execution/task
   */
   logEvent.matchOnAttribute("subType", "TaskLogEvent")
+    .filter(filterDebug)
```
havocp: is the debug filtering still necessary with the DOM cached?

havocp:
This fix looks like the right solution to me. It's a focused change that addresses the exact problem (dom recreation).

henrikengstrom:
LGTM. Do we want to still use filtering of debug though? I tried it without any filtering and it seems pretty fast to me without the filtering too.

ui/app/assets/services/sbt/logs.js
Warry: It mechanically increases the scrollback, while the debug functionality is really supposed to be off 99% of the time.

henrikengstrom:
Okay, merging this one now and we can amend the filtering in case we want to later.

PR #897, opened 2015-02-09
Upgraded to Play 2.3.8-M3 and sbt-rc 1.0-M4

corruptmemory:
Warning, the following integration tests failed for me locally:

```
[error] (activator-offline-tests/*:integrationTests) Offline tests were unsuccessful: reactive-maps, minimal-akka-scala-seed, play-scala, minimal-akka-java-seed, reactive-stocks, hello-akka, minimal-scala, hello-scala, hello-slick-2.1, play-java
```

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -12,12 +12,12 @@ object Dependencies {
   val luceneVersion = "4.2.1"
 
   val templateCacheVersion = "1.0-a0afb008ea619bf9d87dc010156cddffa8a6f880"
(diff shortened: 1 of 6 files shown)


havocp:
PickleToJsonHack can be removed now because toJsonString is now public

havocp:
offline tests should only pass when using henrik's test template for the moment, because we have deliberately not updated the real templates to our snapshot versions

havocp:
we should probably make .travis.yml config use henrik's test template

PR #899, opened 2015-02-09
Protects from potential NP when position info is not available + shows w...

henrikengstrom:
...arnings as well as errors when compiling.

Not sure if we want to show _warnings_ and link to them in the same manner as _errors_?
An idea could be to add toggle functionality in the build tab to enable users to control it?

With this PR we now handle errors and warnings in consistent manner. Before this cleanup a warning was shown as a compilation error in some parts of the UI and hidden in others.

Let me know what you think.

### DIFF
--- a/ui/app/assets/plugins/build/tasks/tasks.html
+++ b/ui/app/assets/plugins/build/tasks/tasks.html
@@ -14,10 +14,16 @@
             <td class="command" data-bind="text: command"></td>
             <td class="data">
             <!-- ko if: finished() -->
-              <!-- ko if: compilationErrors.length -->
+              <!-- ko if: compilationErrors.length && (compilationErrors[0].severity === "Error") -->
               <span class="compilationErrors">
                 <span class="compilationErrorsNumber" data-bind="text: compilationErrors.length"></span>
                 compilation error(s)
+              </span>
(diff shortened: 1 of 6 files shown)


corruptmemory:
LGTM, but I'll wait for other opinions to weigh in.

ui/app/assets/plugins/build/tasks/tasks.html
```diff
@@ -14,12 +14,18 @@
             <td class="command" data-bind="text: command"></td>
             <td class="data">
             <!-- ko if: finished() -->
-              <!-- ko if: compilationErrors.length -->
+              <!-- ko if: compilationErrors.length && (compilationErrors[0].severity === "Error") -->
```
havocp: is there a guarantee that errors are before warnings in the list?

havocp:
this seems like an improvement to be consistent, but surely we should show both errors and warnings, showing them differently though?

ui/app/assets/plugins/build/tasks/tasks.html
henrikengstrom: When there is a compilation error I don't even see the warnings in the log output. Not sure if they are even sent back from the server?

PR #902, opened 2015-02-10
bump to Play 2.3.8-M4 and dependencies

havocp:
This will probably fail to build at first because M4 isn't up yet, but hit retry on Travis once it is.


corruptmemory:
I think you missed: https://github.com/typesafehub/activator/blob/wip/havocp-play-m4/project/play.sbt#L3

corruptmemory:
LGTM

PR #903, opened 2015-02-10
Add tooltip when disabling buttons

Warry:


Warry:
screen shot 2015-02-10 at 14 10 55

ui/app/assets/services/sbt/events.js
```diff
@@ -30,6 +30,39 @@ define([
...
   /**
+  Full text application status
+  */
+  var appStatus = ko.computed(function() {
+    if(!tasks.buildReady()){
+      return { id: "buildFailed", label: "Build loading has failed", url: "#build/tasks" }
+    } else if(tasks.compilationErrors().length){
+      var ers = tasks.compilationErrors();
+      // Goto first compile error
+      var url = "#code"+ fs.relative(ers[0].position.sourcePath)+":"+ers[0].position.line;
```
havocp: position and/or sourcePath can be null for "global" warnings, see henrik's patch

ui/app/assets/widgets/appStatus/appStatus.html
```diff
@@ -1,4 +1,4 @@
-<div id="appStatus">
+<div id="appStatus" data-bind="tooltip: whyDisabled">
```
havocp: should this also include the "not disabled" statuses like "running project"?

henrikengstrom: The "run" button switches label to become "stop" when the project is running. That should be enough shouldn't it?

havocp: I thought this tooltip was on all four buttons, giving general app status.

henrikengstrom: Sorry for being unclear. It is for all buttons. Some of them switch when invoked though to clarify what is going on, e.g. "Compile" -> "Stop", "Run" -> "Stop". It is likely that when a user clicks on "compile" or "run" they are hanging around on that tab to see the outcome rather than jumping around to other tabs and clicking buttons. This is what I meant. 
If you think it still makes sense to add a more general tooltip functionality then I won't stand in the way for that. I just think that it won't come to use as much.

Warry: There are other tooltips when not disabled (not disabled can mean: steady, active or working)

havocp: oh I see, if we are not disabled then we have individual button tooltips, when we are disabled a tooltip on all four little icons. Is it guaranteed that we never show both?

Warry: Now we are.

PR #904, opened 2015-02-10
Show active task even if it isn't a known one

havocp:
This makes things feel a lot nicer if I type "update"
or something like that in the search box.

Also, clarified "Activator is running smoothly" to
"no errors or activity right now" which I think makes
more sense. The issue is whether the app has errors,
not how activator is.


Warry:
LGTM

PR #906, opened 2015-02-10
Change when/how we clear out errors and test results

havocp:
Previously, if you did a compile and then some unrelated task,
we would clear out your errors. Instead, track whenever a compileIncremental
result changes (which could be part of some other task), and when it does
the errors from the execution in progress will "take over" as the current
errors.

Also, always prefer errors and tests from the execution in progress
over errors from prior ones; this means that errors will stream in,
rather than all showing up at the end.

This introduces an error log message about "can't serialize the result of compileIncremental," so I have to go remove that println from sbt-remote-control I think. but we don't have to block on that.

### DIFF
--- a/ui/app/assets/services/sbt/tasks.js
+++ b/ui/app/assets/services/sbt/tasks.js
@@ -234,7 +234,13 @@ define([
       // we want to be in the by-id hash before we notify
       // on the tasks array
       tasksById[task.taskId] = task;
-      executionsById[task.executionId].tasks[task.taskId] = task;
+      execution.tasks[task.taskId] = task;
+
+      if (task.key === "compile" || task.key === "compileIncremental") {
+        // for most executions we'll get "compile" AND "compileIncremental",
+        // so this has to be idempotent
+        execution.changedCompileResult = true;
+      }
     } else {
       debug && console.log("Ignoring task for unknown execution " + message.event.executionId)
     }
@@ -495,12 +501,6 @@ define([
     }
   });
 
-  valueChanged.matchOnAttribute('key', 'compileIncremental').each(function(message) {
-    var current = workingTasks.current();
-    if (current)
-      current.changedCompileResult = true;
-  });
-
   // Inspect-related (sbt-echo) observables.
   //
   // FIXME these need to be tracked separately for each project.
@@ -629,8 +629,7 @@ define([
 
     // Data produced:
     self.tasks          = {};
-    // true if during the execution we got a notify on compileIncremental.
-    // note that this becomes true AFTER we get all the errors.
+    // true if during the execution we see a compile task
     self.changedCompileResult = false;
     // if this is non-empty, then changedCompileResult ought to end up true...
     self.compilationErrors  = ko.observableArray([]);
--- a/ui/app/assets/services/sbt/tasks.js
+++ b/ui/app/assets/services/sbt/tasks.js
@@ -50,6 +50,9 @@ define([
     test:     ko.observable(false)
   };
 
+  var mostRecentWithCompilationErrors = ko.observable(null);
+  var mostRecentWithTestResults = ko.observable(null);
+
   /**
   Stream Events
   */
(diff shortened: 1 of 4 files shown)


ui/app/assets/services/sbt/tasks.js
```diff
@@ -205,7 +234,13 @@ define([
       // we want to be in the by-id hash before we notify
       // on the tasks array
       tasksById[task.taskId] = task;
-      executionsById[task.executionId].tasks[task.taskId] = task;
+      execution.tasks[task.taskId] = task;
+
+      if (task.key === "compile" || task.key === "compileIncremental") {
+        // for most executions we'll get "compile" AND "compileIncremental",
+        // so this has to be idempotent
+        execution.changedCompileResult = true;
```
henrikengstrom: We are not setting this to false for all other tasks, are we? So it will stay true once we've received one ``compile`` or ``compileIncremental``.

havocp: that's the correct behavior; if an execution contains a compile (say, discoveredMainClasses or run would contain one), then we want to reset errors and use any errors that happened during that execution.

henrikengstrom: Not sure I understand what executions we can expect here? What I am saying is that we are trying to guard the code from some other executions here: https://github.com/typesafehub/activator/pull/906/files#diff-a0280aab319302b62a5198133daac50dR130
Once we have received a task with ``compile`` and/or ``compileIncremental`` this guard will always be true, i.e. not protect the counter from any of the executions we guarded it from in the first place. 
But as I say, you have a better idea of what we can expect when so as long as this makes sense to you I am good with it.

ui/app/assets/services/sbt/tasks.js
```diff
@@ -70,13 +73,39 @@ define([
...
+      return [];
+    }
+  });
   var testErrors = ko.computed(function() {
     return testResults().filter(function(t) {
       return t.outcome === "failed";
     });
   });
-  var compilationErrors = ko.observableArray([]);
+  var compilationErrors = ko.computed(function() {
```
henrikengstrom: Changing this from an array to function stopped this code from working:
https://github.com/typesafehub/activator/blob/wip/havocp-clearing-errors/ui/app/assets/plugins/build/tasks/tasks.html#L17-27

ui/app/assets/services/sbt/tasks.js
havocp: All executions should go through here. An execution contains tasks, and the "execution" var here is a single execution (like a single command from the search box). So once an execution has a compile task in it, then it's a compile-affecting execution that should take over and become the current error list. If the execution has nothing to do with compile (say it's "update" and doesn't cause a compile), then it will have an empty compile error list but we want to ignore that and keep any compile errors we have. But if the execution involves a compile (say discoveredMainClasses), then its compile error list is meaningful (whether empty or nonempty).

Read over and be sure it makes sense to you, this is going to be important to understand to clean up some of this code. there's a lot more cleanup to do here I think.

The events we will get will be something like:
  * ExecutionWaiting("discoveredMainClasses")
  * ExecutionStarting("discoveredMainClasses") 
  * all sorts of TaskStarted, including TaskStarted("compile")
  * possible 0-N CompilationFailure events
  * TaskFinished for each task started
  * ExecutionFinished("discoveredMainClasses")

When we look at execution.command == "compile" that's wrong because we want to know if the execution contains a compile, not whether compile is the 'target task' at the end of the execution. A problem that complicates matters is that we don't currently get an event with all the tasks in the execution up front; we only get TaskStarted as the tasks actually begin running. So we may need more events from sbt-remote-control before we're completely happy with this.

henrikengstrom:
LGTM apart from the comments/questions above.
Travis is all good with it. Should just merge while Travis is green!!!

ui/app/assets/services/sbt/tasks.js
havocp: should be fixed

ui/app/assets/services/sbt/tasks.js
henrikengstrom: Thanks for the clarification. It's is all complicated by tasks that are not a direct compile command can contain a compile anyhow and contain compilation result. It makes sense that ``execution.command === "compile"`` is not good enough with your description of what is going on here above.

I guess another thing that is confusing me is that there can be an execution that has a ``execution.complationErrors()`` that should not affect the count. This is why the guard has been added here, isn't it?
https://github.com/typesafehub/activator/pull/906/files#diff-a0280aab319302b62a5198133daac50dR130

henrikengstrom:
LGTM (again)

ui/app/assets/services/sbt/tasks.js
havocp: right, that's just to avoid special-casing null or something for execution.compilationErrors. if there's no compile there shouldn't be any compilation errors. (Though nothing is really stopping some random task from sending a compilation error event if it wanted to.)

PR #908, opened 2015-02-10
Forward DetachedEvent to JavaScript

havocp:
This has no practical effect for now except to prevent a MatchError,
but will be needed to get RP events later.


henrikengstrom:
LGTM

PR #910, opened 2015-02-11
Don't show debug logs when opening a new project

havocp:
The log view here isn't scalable, and debug logs for ivy
resolution are ~15k lines vs ~500 lines for non-debug.
This makes things several times faster.


corruptmemory:
LGTM

PR #911, opened 2015-02-11
Delete some dead code in JsonHelper.scala that caused compiler warnings

havocp:


corruptmemory:
Nice, negative additions! :-) 

LGTM.

PR #913, opened 2015-02-11
Uses template instead name to retrieve the metadata for a project.

henrikengstrom:


ui/app/assets/services/tutorial.js
```diff
@@ -42,8 +42,7 @@ define([
     return index() === table().length-1;
   });
 
-  // TODO = provide JSON route, for meta-datas
-  $.getJSON("/api/templates/"+serverAppModel.name+"/meta", function(data){
+  $.getJSON("/api/templates/" + serverAppModel.template + "/meta", function(data){
```
havocp: serverAppModel.template is an Option (https://github.com/typesafehub/activator/blob/master/ui/app/controllers/Application.scala#L28 ) which I believe means it can be either undefined or null here, if an app wasn't created from a template or somebody deleted the uuid from build.properties. It would probably be good to handle that explicitly rather than relying on a 404 for "/api/templates/undefined/meta"

havocp:
LGTM other than handling the no-template-id case.

ui/app/assets/services/tutorial.js
henrikengstrom: What do you mean by handle it explicitly? There is not much we can do if it's not present. We could warn the user that the information needed to retrieve the template metadata is not available?

It becomes a "undefined" which results in a 404 as you say. This means that the template info in the popup (bubble) won't be show.

henrikengstrom: I could of course prevent the request from happening at all if the data is missing...

havocp: yeah, I just meant something like `if (serverAppModel.template) getJson() else { do something, or nothing perhaps }` - that way the code is explicitly making clear that we may not get the metadata.

henrikengstrom: Got it.

Warry:
LGTM too

PR #914, opened 2015-02-11
Add stream and spark to tag list

Warry:


ui/app/controllers/Application.scala
```diff
@@ -90,7 +90,7 @@ object Application extends Controller {
       templates = featured,
       otherTemplateCount = tempSeq.length,
       recentApps = getAppsThatExist(config.applications),
-      tags = Seq("reactive", "scala", "java", "java8", "starter", "akka", "play", "slick", "spray", "angular", "javascript", "database", "websocket"))
+      tags = Seq("reactive", "scala", "java", "java8", "starter", "akka", "play", "slick", "spray", "stream", "spark", "angular", "javascript", "database", "websocket"))
```
dotta: nitpicking, but `streams` might be better (because of reactive streams, akka streams)

PR #915, opened 2015-02-11
Prevent registration of existing mime detector

henrikengstrom:


havocp:
that turned out to be so easy I now feel lame we left it broken for so long...

PR #918, opened 2015-02-11
Synchronize our use of the MIME library

havocp:
Our "only set detector if it's null" code is certainly not threadsafe,
and possibly some NPEs we are seeing are also caused by concurrent
usage, though that's pure speculation.


corruptmemory:
Breaking out the bug guns :-).  LGTM.

PR #919, opened 2015-02-11
Updated echo and sbt-echo

corruptmemory:
* sbt-echo properly uses sbt-core-next
* echo updated to latest Play (2.3.8-M7)

### DIFF
--- a/sbt-echo/akka/src/main/scala/com/typesafe/sbt/SbtEcho.scala
+++ b/sbt-echo/akka/src/main/scala/com/typesafe/sbt/SbtEcho.scala
@@ -6,11 +6,13 @@ package com.typesafe.sbt
 import sbt._
 import sbt.Keys._
 import java.net.URI
+import sbt.plugins.BackgroundRunPlugin
+import sbt.BackgroundJobServiceKeys
 
 object SbtEcho extends AutoPlugin {
   import echo.EchoRun._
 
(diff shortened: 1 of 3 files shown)


corruptmemory:
This does fix the problem of opening projects from the console.  sbt-echo was waaaaaaaay out of date, still referring to `UIKeys` and all.  Honestly, I have no idea how it even worked at all.  This code didn't even compile.

I published a 0.1.10 version of stuff already.

sbt-echo/akka/src/main/scala/com/typesafe/sbt/SbtEcho.scala
```diff
@@ -6,11 +6,13 @@ package com.typesafe.sbt
...
 import sbt.Keys._
 import java.net.URI
+import sbt.plugins.BackgroundRunPlugin
+import sbt.BackgroundJobServiceKeys
 
 object SbtEcho extends AutoPlugin {
   import echo.EchoRun._
 
-  val EchoVersion = "0.1.9"
+  val EchoVersion = "0.1.10"
```
havocp: doesn't have to be (shouldn't be) in this PR but can this just point to Dependencies.echoPluginVersion ?

corruptmemory: Sure, no objections from me. :-)

PR #920, opened 2015-02-12
Rework how we set up repos - we needed to handle several more cases

havocp:
Large comment in RepositoryConfig.scala documents the particulars.

### DIFF
--- a/launcher/src/main/scala/activator/RepositoryConfig.scala
+++ b/launcher/src/main/scala/activator/RepositoryConfig.scala
@@ -17,22 +17,30 @@ import java.io.File
 // Important background:
 //  - there's a default repository configuration embedded in the
 //    launcher, which we generate in project/Packaging.scala
-//  - here we are creating ~/.sbt/repositories
+//  - the launcher also configures a repository-config, which
+//    is an override file for the repositories section in the
+//    launcher
+//  - in this source file we are creating ~/.sbt/repositories
 //  - if ~/.sbt/repositories exists, sbt will IGNORE the entire
 //    embedded config in the launcher, not merge with it.
 //    ~/.sbt/repositories entirely replaces the embedded config.
-//  - activator.home is set to the unpacked activator zip
-//    directory by the activator/activator.bat wrapper scripts
+//    This is how the repository-config is set up in
+//    Packaging.scala.
+//  - activator.home is set to the activator/activator.bat wrapper
+//    scripts' location
 //  - ${activator.home}/repository would be our offline repo
+//    for the "fat" zip
 //  - in the "minimal" zip, ${activator.home}/repository doesn't
 //    exist
+//  - when the activator scripts are copied into an app,
+//    ${activator.home}/repository doesn't exist
 //  - activator.local.repository is a user-configurable override
 //    that replaces ${activator.home}/repository (you would
 //    specify this on the activator command line or in
 //    ~/.sbt/jvmargs)
 //  - if we don't have activator.home we seem to use
 //    ${user.home}/.activator, but with our wrapper scripts it
-//    isn't clear that can ever happen.
+//    isn't clear that can ever happen. (TODO clean up?)
 //
 // With that background, we are trying to handle these cases:
 //  - if you've never run activator and run it the first time,
@@ -41,8 +49,13 @@ import java.io.File
 //  - the second time you run activator, both launcher and sbt
 //    server would use ~/.sbt/repositories
 //  - if you upgrade to a new Activator (which probably moves
-//    activator.home), it uses ~/.sbt/repositories and ignores
-//    the embedded launcher config
+//    activator.home), the first time you run it, it uses
+//    ~/.sbt/repositories which will configure both the previous
+//    version of activator's repository and also the new version's
+//    repository due to activator.home
+//  - if you upgrade, the second time you run it we'll have edited
+//    ~/.sbt/repositories and it will only use the new version's
+//    repo
 //
 // We write two repos to ~/.sbt/repositories; one is hardcoded
 // to the activator path when we created ~/.sbt/repositories,
@@ -60,7 +73,9 @@ import java.io.File
 // that start with activator-.
 //
 // We also only edit ~/.sbt/repositories if we are a "fat" zip,
-// that is if ${activator.home}/repository exists.
+// that is if ${activator.home}/repository exists. If you run
+// an activator from inside an app or from the minimal zip,
+// we use the last "fat" repository you have used.
 object RepositoryConfig {
   private val repositoriesSectionName = "repositories"
 
--- a/launcher/src/main/scala/activator/RepositoryConfig.scala
+++ b/launcher/src/main/scala/activator/RepositoryConfig.scala
@@ -7,6 +7,60 @@ import scala.util.control.NonFatal
 import activator.properties.ActivatorProperties._
 import java.io.File
 
+// Here's where we specify our repositories configuration.  The
+// goal is to add the offline repository that comes in our zip
+// file to Activator itself AND to sbt server.
+//
+// Note: as with any comment this one can be wrong. This is how
+// someone THOUGHT it worked at one point. Trust but verify!
+//
+// Important background:
+//  - there's a default repository configuration embedded in the
+//    launcher, which we generate in project/Packaging.scala
+//  - here we are creating ~/.sbt/repositories
+//  - if ~/.sbt/repositories exists, sbt will IGNORE the entire
+//    embedded config in the launcher, not merge with it.
+//    ~/.sbt/repositories entirely replaces the embedded config.
+//  - activator.home is set to the unpacked activator zip
+//    directory by the activator/activator.bat wrapper scripts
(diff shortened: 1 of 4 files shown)


havocp:
Oh, I stopped adding comments to ~/.sbt/repositories because the code failed to remove them again so they accumulated.

havocp:
@jsuereth should probably see if I screwed this up.

corruptmemory:
Generally LGTM.

launcher/src/main/scala/activator/RepositoryConfig.scala
```diff
@@ -7,6 +7,60 @@ import scala.util.control.NonFatal
...
+// file to Activator itself AND to sbt server.
+//
+// Note: as with any comment this one can be wrong. This is how
+// someone THOUGHT it worked at one point. Trust but verify!
+//
+// Important background:
+//  - there's a default repository configuration embedded in the
+//    launcher, which we generate in project/Packaging.scala
+//  - here we are creating ~/.sbt/repositories
+//  - if ~/.sbt/repositories exists, sbt will IGNORE the entire
```
jsuereth: Note: it's actually nuanced.   

1. THe launcher config you use *can* specify a repositories override file (you don't have to).  
2. If the file specified exists, then it is used to completely replace the repositories section in the embedded launcher config.
3. The default setting first looks for a setting called sbt.repositories.config (or some such) and then looks at ~/.sbt/repositories.  You'll see this in your own embedded laucnher properties.

launcher/src/main/scala/activator/RepositoryConfig.scala
```diff
@@ -7,6 +7,60 @@ import scala.util.control.NonFatal
...
+//    embedded config in the launcher, not merge with it.
+//    ~/.sbt/repositories entirely replaces the embedded config.
+//  - activator.home is set to the unpacked activator zip
```
jsuereth: Again, it's nuanced. activator.home points to the directory where the activator.bat/activator (bash) scripts are run.  THis may or may not be an actual activator home, in the instance where the scripts are injected right into a project.

launcher/src/main/scala/activator/RepositoryConfig.scala
```diff
@@ -7,6 +7,60 @@ import scala.util.control.NonFatal
...
+//    directory by the activator/activator.bat wrapper scripts
+//  - ${activator.home}/repository would be our offline repo
+//  - in the "minimal" zip, ${activator.home}/repository doesn't
```
jsuereth: This is also the case when using an activator script injected into a cloned template.

launcher/src/main/scala/activator/RepositoryConfig.scala
```diff
@@ -7,6 +7,60 @@ import scala.util.control.NonFatal
...
+//    exist
+//  - activator.local.repository is a user-configurable override
+//    that replaces ${activator.home}/repository (you would
+//    specify this on the activator command line or in
+//    ~/.sbt/jvmargs)
+//  - if we don't have activator.home we seem to use
```
jsuereth: this seems more of a bug of attrition rather than by design...

launcher/src/main/scala/activator/RepositoryConfig.scala
```diff
@@ -7,6 +7,60 @@ import scala.util.control.NonFatal
...
+//    ${user.home}/.activator, but with our wrapper scripts it
+//    isn't clear that can ever happen.
+//
+// With that background, we are trying to handle these cases:
+//  - if you've never run activator and run it the first time,
```
jsuereth: A case to think about:  THe user already has a ~/.sbt/repositories file and proxy.   You may want to issue some kind of warning if activator is not found about how they need to proxy the activator repository for this to work.  (Which I think they'd have to anyway for sbt, so maybe it's a moot point).

launcher/src/main/scala/activator/RepositoryConfig.scala
```diff
@@ -7,6 +7,60 @@ import scala.util.control.NonFatal
...
+//    the embedded launcher config
+//
+// We write two repos to ~/.sbt/repositories; one is hardcoded
+// to the activator path when we created ~/.sbt/repositories,
+// and this is intended to affect sbt server. The other honors
+// activator.home which is set by the wrapper script, so this
+// one is intended to work right with the Activator launcher
+// (which may be a newer version).
+//
+// In the "average" case (the second and subsequent times you
```
jsuereth: What about the "upgrade" case.  Do you move the default ofline repository to point at the new hardcoded path or does it remain against the old one?

launcher/src/main/scala/activator/RepositoryConfig.scala
```diff
@@ -35,28 +89,41 @@ object RepositoryConfig {
...
-          old.lines.filterNot(_.trim.startsWith("activator-local:")).partition(_.trim == "local")
-        val oldLine = old.lines.find(_.trim.startsWith("activator-local:"))
-        // this check is to avoid adding the comment if we haven't really changed anything
-        if (oldLine != Some(activatorLocalLine))
-          old.copy(lines = before ++ Seq("  # Activator added this local repository",
-            activatorLocalLine) ++ after)
-        else
-          old
+        val withActivatorLocal =
+          replaceRepoLine(old, "local", "activator-local", activatorLocalLine)
```
jsuereth: what is this "local" string about?

launcher/src/test/scala/activator/RepositoryConfigTest.scala
```diff
@@ -151,8 +151,8 @@ class RepositoryConfigTest {
   repository-config: ${sbt.repository.config-${sbt.global.base-${user.home}/.sbt}/repositories}
```
jsuereth: FYI - This is where the default repository override file is specified.  Note that it is influenced by `sbt.repository.conf` and `sbt.globa.base` and `user.home`.

ui/app/activator/UIMain.scala
```diff
@@ -94,16 +94,12 @@ class UIMain extends AppMain {
...
-            // Fix UNC path problem on Windows http://www.tomergabel.com/JavaMishandlesUNCPathsOnWindows.aspx
-            var uri: URI = x.url.toURI
-            if (uri.getAuthority != null) uri = new URI(uri.toString.replace("file://", "file:/"))
-            System.out.println("Activator will use extra repository " + x.id + " @ " + uri)
-            (x.id, new java.io.File(uri), Some(x.ivyPattern))
-        }
+      // log which repo we are using
+      configuration.provider.scalaProvider.launcher.appRepositories foreach {
+        case x: xsbti.IvyRepository if (x.id == "activator-local" || x.id == "activator-launcher-local") && (x.url.getProtocol == "file") =>
+          System.out.println("Local repository: " + x.id + " @ " + x.url)
```
jsuereth: you may want to remove the debugging output here...

jsuereth:
So the question I have is about forcing your injected local repo to happen between "local" and "maven-central" or whatever.  I guess this is the way we ensure we do not much in the repositories file if it already exists?


Besides general questions on that, we will need (but maybe not in this PR) to account for the possible different locations of the repository config file, even though I don't think many people change it. I do know sbt-rebel-cut (sbt-extras) can do crazy things with sbt.global.base, but maybe activator doesn't care.

launcher/src/main/scala/activator/RepositoryConfig.scala
havocp: yes - ~/.sbt/repositories will always be edited to point to the most recent activator you ran which had a "repository" directory in its activator.home. So this would ignore "minimal" zips and embedded-in-the-app-directory activator scripts.

launcher/src/main/scala/activator/RepositoryConfig.scala
havocp: inserting the new entries after "local"; I'm doing this because it's how the existing launcher config is set up. I don't know whether order of entries actually matters.

ui/app/activator/UIMain.scala
havocp: I was thinking this might be user-interesting, but maybe it isn't really.

launcher/src/main/scala/activator/RepositoryConfig.scala
havocp: proxy a local repository? if they already have ~/.sbt/repositories the current code will add the activator-local and activator-launcher-local entries.

havocp:
I think it's fine to leave people on their own if they move the repository config file around, at least for the moment.

For "local", this was just trying to keep the order we had before. But if order doesn't matter I suppose there's no need.

havocp:
Do we think this will work as-is? If so we should probably merge, maybe with some TODO

jsuereth:
yeah, LGTM, but the CI is failing

henrikengstrom:
@jsuereth if you check our build history that should not prevent you from merging this PR:
https://travis-ci.org/typesafehub/activator/builds
Unfortunately we haven't had a reliable CI environment for quite some time now :(

PR #921, opened 2015-02-12
Fix layout and sorting in requests

Warry:


henrikengstrom:
screen shot 2015-02-12 at 08 18 57

Some feedback:
1. would be nice if the arrow was moved to the column which we sort on (as in the actors page)
2. all sorting is being done descending which might make sense in some cases but not for the paths/controllers

viktorklang:
Perhaps use the same time unit in that column?

-- 
Cheers,
√

henrikengstrom:
Tests done in Safari btw...

henrikengstrom:
Tests done in Safari/OSX

A couple of more comments:
1. The "arrow" indicating what is currently being sorted on is not in the correct column (always on Method).
2. Sorting on Request, Response Code and Time seems broken
3. In the drop-down there is a Path to sort on but it is called Request in the table.

henrikengstrom:
This should be removed too: https://github.com/typesafehub/activator/blob/master/ui/app/assets/plugins/run/requests/requests.html#L30

Warry:
@henrikengstrom I rebased

henrikengstrom:
Still not working as expected (from previous comment above): 
2. Sorting on Request, Response Code and Time seems broken
3. In the drop-down there is a Path to sort on but it is called Request in the table.

henrikengstrom:
Also, time formatting is incorrect:
screen shot 2015-02-17 at 09 45 40

ui/app/assets/plugins/run/requests/requests.html
```diff
@@ -27,58 +27,36 @@
...
-          <td class="invocation" data-bind="text: invocationTimeMillis"></td>
-          <td class="time" data-bind="text: startTimeMillis"></td>
-        </tr>
-      </tbody>
-    </table>
+        <table class="table-list">
+          <thead>
+            <tr>
+              <th rowspan="2" data-bind="click: filters.toggleOrdering('method'), attr: {'data-order-dir': filters.isOrdering('method') }" class=""><span>Method</span></th>
+              <th><span>Request</span></th>
```
Warry: To add path ordering: 

```
data-bind="click: filters.toggleOrdering('path'), attr: {'data-order-dir': filters.isOrdering('path') }" 
```

ui/app/assets/plugins/run/requests/requests.js
```diff
@@ -21,29 +21,28 @@ define([
...
-    { value: "path",                 text: "Path" },
-    { value: "httpResponseCode",     text: "Response Code" },
-    { value: "controller",           text: "Controller" },
-    { value: "invocationTimeMillis", text: "Invocation Time" },
-    { value: "startTimeMillis",      text: "Time" }
+    { value: "default",       text: "Invocation Time" },
+    { value: "method",        text: "Method" },
+    { value: "responseCode",  text: "Response Code" },
+    { value: "controller",    text: "Controller" },
+    { value: "time",          text: "Time" }
```
Warry: To add path ordering: 
```
{ value: "path",          text: "Path" }
```

henrikengstrom:
LGTM. 
Will add path sorting in a separate PR.

PR #923, opened 2015-02-12
Changes from an incorrect counter to an indication that there are reques...

henrikengstrom:
...ts available.

This is the way we did it in Console too isn't it? Therefore I don't think this is a regression.
Also better to not show incorrect numbers and there is no time to find out the underlying incorrectness in Echo/Analytics atm.

### DIFF
--- a/ui/app/assets/plugins/run/run.html
+++ b/ui/app/assets/plugins/run/run.html
@@ -97,7 +97,7 @@
       </li>
       <li class="requests" data-bind="isActiveUrl: '#run/requests'">
         <a href="#run/requests">Requests</a>
-        <span class="counter" data-bind="text: stats().data.metadata.playPatternCount, visible: stats().data.metadata.playPatternCount"></span>
+        <span class="counter" data-bind="text: stats().data.metadata.playPatternCount > 0 ? 'Y' : 'N' ,visible: stats().data.metadata.playPatternCount"></span>
       </li>
       <!-- /ko -->
       <!-- ko if: monitoringSolutions.newRelicActivated() -->


ui/app/assets/plugins/run/run.html
```diff
@@ -97,7 +97,7 @@
       </li>
       <li class="requests" data-bind="isActiveUrl: '#run/requests'">
         <a href="#run/requests">Requests</a>
-        <span class="counter" data-bind="text: stats().data.metadata.playPatternCount, visible: stats().data.metadata.playPatternCount"></span>
+        <span class="counter" data-bind="text: stats().data.metadata.playPatternCount > 0 ? 'Y' : 'N' ,visible: stats().data.metadata.playPatternCount"></span>
```
corruptmemory: Is it supposed to be Y/N?  Not a count?

henrikengstrom: Yes, since the counter is broken. We had a "Y" in the last version of Activator didn't we?
If you don't think "Y" is a good idea I vote that we just remove the indicator all together as it is showing an incorrect value.

corruptmemory: Yeah, the counter we're using is not accurate.  And yes Y/N was the last thing we did.

Made a trello card to revisit this at some point: https://trello.com/c/zrOzv3mm

henrikengstrom:
My thinking here was that it's better to have something that is correct, although maybe not exactly what we want going forward, for the upcoming release.

corruptmemory:
LGTM

PR #924, opened 2015-02-12
Automate syncing the sbt-echo version with our build

havocp:
The intent here is that you can change the version in Dependencies.scala and then go into the echo and sbt-echo projects and publish those.


henrikengstrom:
LGTM

PR #925, opened 2015-02-12
Use the sbt-echo plugin version for echo as well

havocp:


project/echo.scala
```diff
@@ -18,7 +18,8 @@ object EchoBuild extends Build {
       settings (LocalTemplateRepo.settings: _*)
       settings(
       Keys.resolvers += typesafeIvyReleases,
-      parallelExecution in GlobalScope := false
+      parallelExecution in GlobalScope := false,
+      version := Dependencies.echoPluginVersion
```
henrikengstrom: I feel a more descriptive name than ``echoPluginVersion`` might be good? It becomes a bit ambiguous now that it has impact on both Echo and sbt-echo. Wdyt?

henrikengstrom:
LGTM - pending the name comment.

havocp:
Done. also I published 0.1.11 in there, so I just left the commit where it was to show when I published.

henrikengstrom:
LGTM

PR #926, opened 2015-02-12
Potential fix for the "hidden Play" problem

corruptmemory:
AspectJ's weaver and tools jar contain a top-level class 'Play' that
conflicts with sbt's automagical import of plugins on .sbt files that
use Typesafe Play.  As it turns out we don't need either one of these
artifacts if we do forked background run all the time.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -5,7 +5,7 @@ object Dependencies {
   // this version is used to publish echo and sbt-echo,
   // and also to set which version of those Activator
   // depends on.
-  val echoVersion = "0.1.11"
+  val echoVersion = "0.1.12"
 
   val sbtVersion = "0.13.8-M3"
   val sbtLibraryVersion = "0.13.8-M3" // for sbtIO on scala 2.11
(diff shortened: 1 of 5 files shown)


project/Dependencies.scala
```diff
@@ -63,8 +63,8 @@ object Dependencies {
   val playSbt13Plugin        =  Defaults.sbtPluginExtra("com.typesafe.play" % "sbt-plugin" % play23Version, "0.13", "2.10")
   val eclipseSbt13Plugin     =  Defaults.sbtPluginExtra("com.typesafe.sbteclipse" % "sbteclipse-plugin" % "2.2.0", "0.13", "2.10")
   val ideaSbt13Plugin        =  Defaults.sbtPluginExtra("com.github.mpeltonen" % "sbt-idea" % "1.5.2", "0.13", "2.10")
-  val echoSbt13Plugin        =  Defaults.sbtPluginExtra("com.typesafe.sbt" % "sbt-echo-play" % echoVersion, "0.13", "2.10")
-  val echoPlaySbt13Plugin    =  Defaults.sbtPluginExtra("com.typesafe.sbt" % "sbt-echo" % echoVersion, "0.13", "2.10")
+  val echoSbt13Plugin        =  Defaults.sbtPluginExtra("com.typesafe.sbt" % "sbt-echo" % echoVersion, "0.13", "2.10")
+  val echoPlaySbt13Plugin    =  Defaults.sbtPluginExtra("com.typesafe.sbt" % "sbt-echo-play" % echoVersion, "0.13", "2.10")
```
havocp: heh, that was amusing

havocp:
This seems 100% fine to me. Activator UI requires forking, and we have no real way to view trace data except Activator UI.

project/Dependencies.scala
```diff
@@ -5,7 +5,7 @@ object Dependencies {
   // this version is used to publish echo and sbt-echo,
   // and also to set which version of those Activator
   // depends on.
-  val echoVersion = "0.1.11"
+  val echoVersion = "0.1.12"
```
henrikengstrom: This requires a new publish of course.

corruptmemory: on it

PR #927, opened 2015-02-12
Removed hard coded version from HTML

henrikengstrom:
We will warn if someone tries to use Inspect on an unsupported version of Akka. Also, never a good idea to have hard coded version lingering about in the HTML...


corruptmemory:
Agreed.  LGTM

PR #929, opened 2015-02-13
Enable url button for play apps only when play server actually started

Warry:


ui/app/assets/services/sbt/tasks.js
```diff
@@ -157,6 +157,11 @@ define([
   }
 
   var playApplicationUrl = ko.observable();
+  var playServerStarted = ko.observable(false);
+  // Watch whenever any task ends, because we can't target specifically play stop
+  workingTasks.run.subscribe(function(v) {
+    if (!v) playServerStarted(false);
+  });
```
havocp: ko.computed can save a lot of spaghetti:
```
var playServerStarted = ko.computed(function() {
  return workingTasks.run() && playApplicationUrl() !== null;
});
```
You'd still have to do `playApplicationUrl(null)` when the run job stops though.

henrikengstrom:
Can't get this to work properly. 
Tested with OSX/Chrome and Reactive Stocks.

To recreate:
1. Run app 
2. URL button becomes visible after a while but is not clickable
3. Stop app
4. Button becomes clickable after a while but since there is no running app the URL doesn't work

henrikengstrom:
Retested and things seem fine now. Must have been something local on my machine that caused this?

ui/app/assets/services/sbt/tasks.js
```diff
@@ -156,7 +156,14 @@ define([
     return false;
   }
 
-  var playApplicationUrl = ko.observable();
+  var playApplicationUrl = ko.observable(null);
+  var playServerStarted = ko.computed(function() {
+    return workingTasks.run() && playApplicationUrl() !== null;
+  });
+  // Watch whenever any task ends, because we can't target specifically play stop
+  workingTasks.run.subscribe(function(v) {
```
havocp: you can target play stop; it should be a BackgroundJobFinished event on the job ID returned from backgroundRun.

PR #931, opened 2015-02-16
Bring back the build error counter

Warry:


havocp:
Why does this grep the logs? I would expect it to be:

`var errorCounter = ko.computed(function() { return compilationErrors().length; });`

That way it's guaranteed to be in sync with the compilationErrors list.

havocp:
Or I guess `errorCounters.build = ko.computed(...)` but you see what I mean

ui/app/assets/services/sbt/logs.js
```diff
@@ -48,6 +50,10 @@ define([
   logEvent.matchOnAttribute("subType", "DetachedLogEvent")
```
havocp: most compile errors should be TaskLogEvent from the compileIncremental task, rather than detached... (are you getting detached error logs? how can I reproduce that?)

Warry:
You can have many errors that do not happen during "compile", like building the project.

havocp:
in that case I think we'll have to somehow rationalize the situation. If we have a counter with number of error-level logs in one place and a list of CompilationFailure events in another place, without trying to make sense out of it, that's just confusing.

havocp:
when d

havocp:
(pushed the wrong button on my phone)

When does this counter reset to 0?

havocp:
Maybe we should not be using the structures CompilationFailure events at all?

havocp:
I looked at the code more and understand one thing now, which is that we are sending CompilationFailure to errorCounters.code and showing it on the Code plugin, and then this grep of the logs for errorCounters.build goes next to the Build plugin. Since you are only grepping "detached" logs, compile errors (and any other task failure) would show up red in the Build logs, but would not increment the build logs counter.

I didn't figure out yet: when would we set errorCounters.build back to 0?

This change seems like an OK stopgap for 1.3.0 (because it will show a counter for something like failure to load the build), as long as we do set back to 0 sometime.

Warry:
@havocp the counter is set to 0 on render of tasks.js

The whole thing went the wrong way as it seems, I'm working on cleaning it up

henrikengstrom:
Build counter always gets a multiple of 3 for every compilation error. This is because the server sends three ``DetachedLogEvent`` with ``m.event.entry.level === "error"``.
We must figure out a better way of handling this before releasing.

havocp:
I added a commit to coalesce adjacent lines. I think what still feels missing to me is that when we start a compile or start a run, we should clear the counter. Otherwise the counter just goes up and up until you click on the tab, especially since I just made TaskLogEvent also bump the counter... we really need a better way to tell which log messages are CompilationFailure-equivalent.

henrikengstrom:
LGTM

PR #932, opened 2015-02-17
Use RP plugin events to decide warning to display

havocp:
Also, add a stub to watch these events during normal operation,
but for now this just logs them.

This doesn't quite work with the current RP plugin because it doesn't throw an error and stop the build from loading when typesafe.properties is missing. However, the new one that's about to come out will do that and then this should work.

I've also stubbed out here (just debug logged) the RP events for use after loading the project. However, I think this has a basic problem that we'll miss these events if we connect to sbt server after the build has already loaded. So, we may have to change this.

### DIFF
--- a/ui/app/assets/widgets/home/home.js
+++ b/ui/app/assets/widgets/home/home.js
@@ -18,18 +18,49 @@ define([
   modals
 ) {
 
-  var stream = websocket.subscribe('subType', 'BuildFailedToLoad');
-  stream.map(function (e) {
-    var warning = $("<article/>").html("<p>If you are creating Typesafe Reactive Platform project you must add a '<code>typesafe.properties</code>' file in the '<code>&lt;template&gt;/project</code>' folder.</p><p>The file must contain your subscription id in the following format '<code>typesafe.subscription=&lt;YOUR ID&gt;</code>'<br/>For more information see: <a href='http://typesafe.com/subscription'>http://typesafe.com/subscription</a><p>")[0];
-    modals.show({
-      shape: "large",
-      title: "Could not load project",
-      body: warning,
-      cancel: "Return",
-      onCancel: function() {
-        $('#working, #open, #new').toggle();
-      }
-    });
+  var failedDueToNoTypesafeProperties = false;
+  var failedDueToNoSubscriptionId = false;
+
+  var sbtEventStream = websocket.subscribe('type','sbt');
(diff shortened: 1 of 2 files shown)


ui/app/assets/widgets/home/home.js
```diff
@@ -18,18 +18,49 @@ define([
...
+        failedDueToNoTypesafeProperties = true;
+      else if (!event.serialized.subscriptionId)
+        failedDueToNoSubscriptionId = true;
+    }
+  });
+
+  subTypeEventStream("BuildFailedToLoad").each(function (e) {
+    var warning = null;
+
+    if (failedDueToNoTypesafeProperties) {
```
henrikengstrom: Will it always fail for one or the other (not both at the same time)?

ui/app/assets/widgets/home/home.js
```diff
@@ -18,18 +18,49 @@ define([
...
-    });
+  var failedDueToNoTypesafeProperties = false;
+  var failedDueToNoSubscriptionId = false;
+
+  var sbtEventStream = websocket.subscribe('type','sbt');
+  var subTypeEventStream = function(subType) {
+    return sbtEventStream.matchOnAttribute('subType',subType);
+  };
+
+  subTypeEventStream("DetachedEvent").each(function(message) {
```
henrikengstrom: Can we always count that the ``DetachedEvent`` arrives before the ``BuildFailedToLoad`` event?

ui/app/assets/widgets/home/home.js
havocp: yeah - in the SubscriptionIdEvent if fileExists=false then the subscriptionId is always None.

ui/app/assets/widgets/home/home.js
havocp: The SubscriptionIdEvent is always before the plugin throws the exception that fails the load, so for that one yes. The SubscriptionLevelEvent and PlatformRelease event are sent async from a thread, so their order is not defined with respect to loading the build.

henrikengstrom:
LGTM apart from comments/questions.

ui/app/assets/widgets/home/home.js
henrikengstrom: If I'd just read the text below I could've figured out the flow...

PR #933, opened 2015-02-17
Adds dependencies needed for offline tests to pass

henrikengstrom:
With this PR the distro becomes a whopping 452MB.
I am sure there is some trimming to be done, but I couldn't find any low hanging fruit.


havocp:
Should travis pass now with this?

I think it's fine to put this in and then we can see if we can shrink it. I'll have a look.

havocp:
Maybe add a commit to this that puts offlineTests back into travis ?

henrikengstrom:
Will do

PR #935, opened 2015-02-17
Removes trailing version information in case of RP dependencies

henrikengstrom:


sbt-echo/akka/src/main/scala/com/typesafe/sbt/echo/EchoRun.scala
```diff
@@ -24,6 +24,8 @@ object EchoRun {
   val EchoWeave = config("echo-weave").hide
   val EchoSigar = config("echo-sigar").hide
 
+  final val rpPattern = "-bin-rp-"
+
```
havocp: josh said -bin- is a general convention not limited to rp, so maybe just match on the -bin- part.

sbt-echo/play/src/main/scala/com/typesafe/sbt/echo/EchoPlayRun.scala
```diff
@@ -24,9 +24,13 @@ object EchoPlayRun {
     }
   }
 
-  def tracePlayDependency(playVersion: String, echoVersion: String): ModuleID =
-    if (playVersion startsWith "2.3.") "com.typesafe.trace" % ("echo-trace-play-" + playVersion) % echoVersion % EchoTraceCompile.name cross CrossVersion.binary
-    else "com.typesafe.trace" % ("echo-trace-play-" + playVersion) % echoVersion % EchoTraceCompile.name cross CrossVersion.Disabled
+  def tracePlayDependency(playVersion: String, echoVersion: String): ModuleID = {
+    val nonRpPlayVersion =
+      if (playVersion.contains(rpPattern)) playVersion.substring(0, playVersion.indexOf(rpPattern))
+      else playVersion
```
havocp: could factor out a stripBinVersion here and in the Akka case

henrikengstrom:
Thanks. Was too lazy initially.

corruptmemory:
This looks right to me.  Letting 'er rip!

PR #940, opened 2015-02-18
Better run handling? - Don't merge yet...

corruptmemory:
Consider this an experiment on how I think the Run/Stop Play URL experience should be like.  It also works with non-Play apps.  Please try out on your own and see if it makes sense.  If so, then we can fix-up whatever bugs linger and merge it.

### DIFF
--- a/ui/app/assets/plugins/run/run.html
+++ b/ui/app/assets/plugins/run/run.html
@@ -14,7 +14,7 @@
             <!-- /ko -->
           </dd>
         </dl>
-        <button data-bind="css: {disabled: sbt.tasks.applicationNotReady}, click: mainRunAction">
+        <button data-bind="css: {disabled: runDisabled}, enable: runEnabled, click: mainRunAction">
           <svg data-bind="svg: '/assets/icons/run.svg'"/>
           <span class="mainAction" data-bind="text: mainRunName"></span>
         </button>
@@ -75,9 +75,9 @@
           </div>
           Reset data on run
         </li>
-        <!-- ko if: sbt.tasks.playApplicationUrl() -->
+        <!-- ko if: displayPlayUrl() -->
         <li class="play-url">
-          <a target="_blank" data-bind="href: sbt.tasks.playApplicationUrl, text: sbt.tasks.playApplicationUrl, css: {'disabled': !sbt.tasks.playServerStarted()}"></a>
+          <a target="_blank" data-bind="href: playUrl, text: playUrl"></a>
         </li>
         <!-- /ko -->
       </ul>
--- a/ui/app/assets/plugins/run/run.js
+++ b/ui/app/assets/plugins/run/run.js
@@ -36,7 +36,7 @@ define([
   }
   var mainRunAction = function() {
     if (sbt.tasks.pendingTasks.run()){
-      sbt.tasks.actions.kill("run");
+      sbt.tasks.actions.stopRun();
     } else {
       if (sbt.app.settings.automaticResetInspect()) {
         connection.reset();
@@ -45,7 +45,29 @@ define([
     }
   }
   var mainRunName = ko.computed(function() {
-    return sbt.tasks.pendingTasks.run()?"Stop":"Run";
+    if (sbt.tasks.pendingTasks.run()) {
+      return sbt.tasks.pendingTasks.stoppingRun() ? "Stopping" : "Stop";
+    } else {
+      return "Run";
+    }
+  });
+
+  var runEnabled = ko.computed(function() {
+    if (sbt.tasks.applicationReady()) {
+      return !sbt.tasks.pendingTasks.stoppingRun();
+    } else {
+      return false;
+    }
+  });
+
+  var runDisabled = ko.computed(function() { return !runEnabled(); });
+  var playUrl = sbt.tasks.playApplicationUrl;
+  var displayPlayUrl = ko.computed(function() {
+    if ((playUrl() !== null) && !sbt.tasks.pendingTasks.stoppingRun() && sbt.tasks.pendingTasks.run()) {
+      return true;
+    } else {
+      return false;
+    }
   });
 
   var toggleInspect = function() {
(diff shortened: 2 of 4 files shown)


ui/app/assets/plugins/run/run.html
```diff
@@ -75,9 +75,9 @@
           </div>
           Reset data on run
         </li>
-        <!-- ko if: sbt.tasks.playApplicationUrl() -->
+        <!-- ko if: displayPlayUrl() -->
```
havocp: +1 to this and to runDisabled/enabled above, gets the logic out of the HTML

ui/app/assets/plugins/run/run.js
```diff
@@ -44,8 +47,30 @@ define([
       sbt.tasks.actions.run();
     }
   }
-  var mainRunName = ko.computed(function() {
-    return sbt.tasks.pendingTasks.run()?"Stop":"Run";
+  var mainRunName = ko.observable("Run");
+  var runEnabled = ko.observable(false);
```
havocp: if this could be ko.computed, I think it would clarify things.

ui/app/assets/plugins/run/run.js
```diff
@@ -44,8 +47,30 @@ define([
...
+  var runDisabled = ko.computed(function() { return !runEnabled(); });
+  var displayPlayUrl = ko.observable(false);
```
havocp: same for this, ko.computed would make it into a clear specification of _when_ we display the play URL

ui/app/assets/plugins/run/run.js
```diff
@@ -35,8 +35,11 @@ define([
     sbt.tasks.requestExecution(cmd);
   }
   var mainRunAction = function() {
+    displayPlayUrl(false);
     if (sbt.tasks.pendingTasks.run()){
       sbt.tasks.actions.kill("run");
+      runEnabled(false);
+      mainRunName("Stopping");
```
havocp: if we make the kill action set a flag like attemptingRunKill, then runEnabled and mainRunName can be ko.computed using that instead of side-effected here.

ui/app/assets/plugins/run/run.js
```diff
@@ -35,8 +35,11 @@ define([
     sbt.tasks.requestExecution(cmd);
   }
   var mainRunAction = function() {
+    displayPlayUrl(false);
```
havocp: this could be ko.computed by storing the Play url associated with the job that was current when we got the PlayStartedEvent.  Then in ko.computed you see if the job is still alive, and if it isn't, you no longer display the url. i.e. we depend on two things 1) the URL and 2) the list of live jobs. To be fancier, could also bring in attemptingRunKill (below) into the computation.

ui/app/assets/plugins/run/run.js
```diff
@@ -44,8 +47,30 @@ define([
...
+  var displayPlayUrl = ko.observable(false);
+  var playUrl = ko.observable("");
+
+  sbt.tasks.pendingTasks.run.subscribe(function (v) {
+    if (v) {
+      mainRunName("Stop");
+    } else {
+      mainRunName("Run");
+      runEnabled(true);
+    }
```
havocp: using ko.computed with pendingTasks.run and attemptingRunKill I think this whole subscription could be dropped

ui/app/assets/plugins/run/run.js
```diff
@@ -44,8 +47,30 @@ define([
...
+
+  sbt.tasks.applicationReady.subscribe(function (v) {
+    runEnabled(v);
+  });
+
+  sbt.tasks.playApplicationUrl.subscribe(function (v) {
+    if (v !== null) {
+      playUrl(v);
+      displayPlayUrl(true);
+    }
```
havocp: same here this is all computable using a pure function. side effects bad :-)

corruptmemory: LOL!  Except that everything I see on my screen is the result of side-effects! :-)

corruptmemory:
ok, the suggested changes are fine, but more to the point: is this more desirable behavior?

havocp:
I _think_ the gist of your change is to add a third "stopping" state in addition to started or stopped, and that does make sense to me.

ui/app/assets/plugins/run/run.js
havocp: Getting sbt events or user interaction events into observables is inherently a side-effect-based process, but any observable that's derived from those "base" observables can be ko.computed using a pure function. Then the pure function acts as a specification that says "here is how our UI state is computed based on sbt's state" - something that's quite unclear - to me anyway - at the moment.

corruptmemory: Sorry.  My last comment was meant as a joking remark.  Understood, you just want to see the constraints explicitly represented.  Working on an update now.

corruptmemory:
Updated, please have a look.

ui/app/assets/services/sbt/tasks.js
```diff
@@ -735,6 +739,12 @@ define([
     }
   }
 
+  // Helper for stopping run
+  function stop() {
```
havocp: should this be called stopRun ? seems a bit ambiguous

ui/app/assets/services/sbt/tasks.js
```diff
@@ -47,6 +47,7 @@ define([
   var pendingTasks = {
     compile:  ko.observable(false),
     run:      ko.observable(false),
+    stopping: ko.observable(false),
```
havocp: similar I guess - stoppingRun?

havocp:
This looks pretty good to me. Even if it isn't the perfect behavior yet, it's much clearer what the behavior *is*.

corruptmemory:
HA! Sure.  The constraints are now explicit.

havocp:
maybe squash this so it has one commit with a better commit message? then I suppose we should see what @maxime thinks in the morning?

corruptmemory:
Squashed.

Warry:
LGTM

corruptmemory:
Any reason not to merge?

corruptmemory:
Pulling the trigger :-)

PR #941, opened 2015-02-18
Bump to sbt 0.13.8-M5 and sbt-rc that uses it.

havocp:


henrikengstrom:
You're too quick. I was about to do the same (but still waiting for compile to finish...).
LGTM

PR #943, opened 2015-02-19
Always use CrossVersion.binary for akka tracing

havocp:
akka 2.0 and 2.1 aren't used anymore, and the special-case
for a version with a "-" in it broke us on reactive platform.


henrikengstrom:
LGTM!

PR #945, opened 2015-02-19
Improve app status display a bit

havocp:
This makes things feel more informative and accurate.

 * add statuses for build loading / failed to load
 * make "disconnected from backend" take priority
   over all other statuses since things are broken then
 * add status for "waiting to connect to sbt server"
 * put compile in front of applicationNotReady,
   otherwise we could never display compile

There's more we could do here but showing the "loading sbt configuration"
in particular I think makes it clearer what's happening.

Adding an "updating dependencies" status would be very helpful
perhaps but is more work.


corruptmemory:
LGTM

PR #947, opened 2015-02-19
Handle failure to load build during initial analysis

havocp:
Before, we were throwing you back to /home which sucked.
Instead, pick a name based on the directory (we had code
for that if the project just lacked a name setting, already).
If the build fails, show a dialog and require clicking Proceed
before we open the app.


henrikengstrom:
LGTM

PR #948, opened 2015-02-20
Cache hash 8c5a99a36bbf3cd4d6f58dcc2c09ae4e166c7d78 and offline updates

havocp:


havocp:
This is not everything we could probably clean up by a long shot, but removing play-geojson eliminates a whole extra copy of play, so that's a nice start.

henrikengstrom:
LGTM

PR #950, opened 2015-02-20
Removes unused/not needed jars

henrikengstrom:
Ok, this is kind of messed up as a PR since I moved a lot of dependencies around as well as removed some... 

It removes the capability to run Inspect in offline mode (which I am fairly certain will not work regardless of if we pack the JARs below or not). 

JARs truly removed are:
* "org.specs2" % "specs2-analysis_2.11" % "2.3.12"
* "org.specs2" % "specs2-matcher-extra_2.11" % "2.3.12"
* "org.scala-lang" % "scala-compiler" % "2.11.0"
* "com.typesafe.trace" % "echo-trace-akka-2.3.9_2.11" % Dependencies.echoVersion
* "com.typesafe.trace" % "echo-sigar-libs" % Dependencies.echoVersion
* Dependencies.aspectjWeaver

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -202,7 +202,6 @@ object TheActivatorBuild extends Build {
         (Keys.projectID in ref) apply { id => id }
       }).join,
       localRepoArtifacts ++= Seq(
-
         // base dependencies
         "org.scala-sbt" % "sbt" % Dependencies.sbtVersion,
         "org.scala-lang" % "scala-compiler" % Dependencies.sbtPluginScalaVersion,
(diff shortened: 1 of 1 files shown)


henrikengstrom:
Oh, and please run ``offlineTests`` locally to make sure that everything works for you.
I get a 428.1MB distribution after the above has been removed.

corruptmemory:
ok, testing now.  Otherwise LGTM

corruptmemory:
I'm getting a zip: 409M (427968706).  Offline tests pass.

PR #956, opened 2015-02-24
Minor UI fixes

Warry:


Warry:
I added remove news, we can rebase it if we want to keep 'em

PR #957, opened 2015-02-24
IE fix for welcome screen

Warry:


ui/app/assets/plugins/welcome/welcome.js
```diff
@@ -16,16 +16,20 @@ define([
...
+    var style;
+
+    try {
+      // Create the <style> tag
+      style = document.createElement("style");
+      // WebKit hack :(
+      style.appendChild(document.createTextNode(""));
+      // Add the <style> element to the page
+      style.sheet.insertRule(cssStr);
+    } catch(e){}
```
havocp: so when does this throw, always on IE? so the presentation mode checkbox does nothing on IE?

PR #961, opened 2015-02-25
Adds sorting capabilities to error column.

henrikengstrom:


Warry:
So it was JS in the end :)
Pulling it right away

henrikengstrom:
Yeah, sorry for stealing JS related tickets ;)
Actually I could have just changed one line in the backend to achieve the same things, but it made sense to use singular instead of plural for the JS naming so I decided to change that instead.

Warry:
LGTM

PR #963, opened 2015-02-26
Makes echo and sbt-echo use the same version number as Activator.

henrikengstrom:
Stab at using the same version for ``Echo`` and ``sbtEcho`` as ``Activator``. Introduces a ``activatorStableBuild`` variable to use for the plugin references, but I am not sure if this is the way we want to roll? Suggestions are welcome.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -2,10 +2,8 @@ import sbt._
 import Keys._
 
 object Dependencies {
-  // this version is used to publish echo and sbt-echo,
-  // and also to set which version of those Activator
-  // depends on.
-  val echoVersion = "0.1.14"
+  // Used for plugins required by this build
+  val activatorStableVersion = "1.3.0"
 
   val sbtVersion = "0.13.8-M5"
   val sbtLibraryVersion = "0.13.8-M5" // for sbtIO on scala 2.11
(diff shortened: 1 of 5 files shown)


corruptmemory:
So we still have to publish-local or publish to use this stuff?

henrikengstrom:
Yes, that's the current idea. Suggestions are welcome though.

henrikengstrom:
I guess we could make it fully integrated with Activator, i.e. publish it as one unit but in that case I would prefer to incorporate Elucidator into Activator first. Wdyt?

corruptmemory:
LOL!  Well, the suggestion would be to magically make things work without needing to publish more artifacts. :-).  But that's more than a suggestion; that's a "send me the PR to review."

corruptmemory:
LGTM

corruptmemory:
yes, that makes sense.  We'll start with this one.

project/Dependencies.scala
```diff
@@ -2,10 +2,8 @@ import sbt._
 import Keys._
 
 object Dependencies {
-  // this version is used to publish echo and sbt-echo,
-  // and also to set which version of those Activator
-  // depends on.
-  val echoVersion = "0.1.14"
+  // Used for plugins required by this build
+  val activatorStableVersion = "1.3.0"
```
havocp: Doesn't this mean we can never publish new plugins (except if we overwrite them)? If the version is still distinct from the activator version with separate publish, let's call it echoVersion still I think.

project/build.scala
```diff
@@ -38,7 +35,7 @@ object TheActivatorBuild extends Build {
     .noAutoPgp
     .doNotPublish
     aggregate(toReferences(publishedProjects ++
-      Seq(dist, it, localTemplateRepo, offlinetests)): _*)
+      Seq(dist, it, localTemplateRepo, offlinetests, EchoBuild.echo, SbtEchoBuild.sbtEcho)): _*)
```
havocp: this is sort of the meat of the change I suppose, pulling echo into the root aggregate so it compiles and publishes along with everything else. But if we're doing that we have to bump the version whenever we bump the activator version, or we'll overwrite stuff.

(detail here: if echo is published, but isn't in "publishedProjects", that is confusing.)

PR #976, opened 2015-03-02
Add markdown formatted links in descriptions

Warry:


Warry:
Will also need a typesafe.com update, do not merge until then

Warry:
Pretty good to GO (wiki was updated as well)

ui/app/assets/commons/format.js
```diff
@@ -261,6 +261,19 @@ define(function() {
       return (nanos / 1000000).toFixed(decimals);
     };
 
+    Format.prototype.markdownLinks = function(str){
+      // html entities
+      str = str.replace(/&/g, '&amp;')
+        .replace(/</g, '&lt;')
+        .replace(/>/g, '&gt;')
+        .replace(/"/g, '&quot;');
```
havocp: missing single quote here

ui/app/assets/commons/templates.js
```diff
@@ -357,12 +357,14 @@ define(["commons/format"], function(format) {
...
   }
 
-  ko.bindingHandlers.trpify = {
+  ko.bindingHandlers.markdownLinkify = {
     update: function(element, valueAccessor, allBindings, viewModel, bindingContext) {
       var val = valueAccessor();
       var txt = ko.isObservable(val)?val():val;
       $(element).text(txt).html(function(index, text) {
-        return text.replace(/(Typesafe )?Reactive Platform/ig, '<a href="https://typesafe.com/subscription">Typesafe Reactive Platform</a>');
+        text = format.markdownLinks(text);
```
havocp: `var text` ? (it's html though really right? `var html`)

ui/app/assets/commons/format.js
Warry: There is no (need) html replacement for the single quote

ui/app/assets/commons/templates.js
Warry: it's a function argument

ui/app/assets/commons/format.js
havocp: below you place some of the text you've just escaped into an attribute, was my concern. If you have the attribute `href="can't"`, will that confuse anything or are single quotes ignored inside double? According to this answer what you have is fine: http://stackoverflow.com/questions/9187946/escaping-inside-html-tag-attribute-value

Here's a scary post that says IE treats backtick as a quote:
http://wonko.com/post/html-escaping

This says to just escape any non-alphanumeric: https://www.owasp.org/index.php/XSS_%28Cross_Site_Scripting%29_Prevention_Cheat_Sheet#RULE_.232_-_Attribute_Escape_Before_Inserting_Untrusted_Data_into_HTML_Common_Attributes

A possible way to avoid thinking about this whole issue would be to build a dom node with jquery here, instead of escaping and building html as a string.

ui/app/assets/commons/templates.js
havocp: oh, you're assigning to a parameter, that's so naughty I didn't expect it :-P in addition it's passed in as text but then you change it to html but it's still called text

henrikengstrom:
@Warry, this needs more testing or alternative implementation. See @havocp's comment above.

Warry:
screen shot 2015-03-11 at 11 35 40

Warry:
BTW, this should be fine now

henrikengstrom:
LGTM

PR #977, opened 2015-03-09
Being able to manualy resize panel

Warry:


Warry:
Not tried yet on windows

henrikengstrom:
Too good for Windows now are we? :)

Warry:
Now it is, and it (is slow) works !

henrikengstrom:
LGTM

PR #981, opened 2015-03-10
Lengthen timeout for template cache operations

havocp:
Moved it to 4 minutes from 12 seconds, so only truly stuck operations should
time out. We don't want to fail just because of a slow network.


henrikengstrom:
4 minutes ftw!
LGTM

PR #983, opened 2015-03-11
Upgrade SBT web plugins

benmccann:


jsuereth:
Thanks for the patch @benmccann This LGTM. I think maybe the hold up is the offline repo.

cc> @havocp

PR #993, opened 2015-03-31
Parse ANSI codes in log messages and provide a messageHtml field

havocp:
The messageHtml field is pre-escaped of course, and has tags where
the ANSI codes were.

cc @jsuereth to look at ANSI parsing goo, @corruptmemory to look at overall activator changes

### DIFF
--- a/ui/app/activator/AnsiToHtml.scala
+++ b/ui/app/activator/AnsiToHtml.scala
@@ -1,0 +1,77 @@
+package activator
+
+object AnsiToHtml {
+
+  private def ansiToTag(endCode: Int, args: String): String = {
+    endCode match {
+      case 'm' =>
+        if (args == "0") {
+          "</span>"
+        } else if (args.startsWith("3")) {
+          val color = args match {
+            case "30" => "black"
+            case "31" => "red"
+            case "32" => "green"
+            case "33" => "yellow"
+            case "34" => "blue"
+            case "35" => "magenta"
+            case "36" => "cyan"
+            case "37" => "white"
+            case _ => ""
+          }
+          s"""<span style="color: ${color}">"""
+        } else if (args.startsWith("4")) {
+          val color = args match {
+            case "40" => "black"
+            case "41" => "red"
+            case "42" => "green"
+            case "43" => "yellow"
+            case "44" => "blue"
+            case "45" => "magenta"
+            case "46" => "cyan"
+            case "47" => "white"
+            case _ => ""
+          }
+          s"""<span style="background-color: ${color}">"""
+        } else {
+          s"<span>"
+        }
+      case _ => ""
+    }
+  }
+
+  def ansiToHtml(text: String): String = {
+    val builder = new StringBuilder(text.length + 7)
+    var i = 0
+    while (i < text.length) {
+      text.charAt(i) match {
+        case '<' => builder.append("&lt;")
+        case '>' => builder.append("&gt;")
+        case '"' => builder.append("&quot;")
+        case '\'' => builder.append("&#x27;")
+        case '&' => builder.append("&amp;")
+        case 27 if ((i + 1) < text.length) && (text.charAt(i + 1) == '[') =>
+          // any character in 64-126 ends the sequence (which is mostly
+          // letters, notably not digits)
+          var j = i + 2
+          while (j < text.length && (text.charAt(j) < 64 || text.charAt(j) > 126)) {
+            j += 1;
+          }
+          val (endCode, args): (Int, String) =
+            if (j < text.length) {
+              (text.charAt(j), text.substring(i + 2, j))
+            } else {
+              (0, "")
+            }
+          i = j // note that this will still get another +1 below
+          builder.append(ansiToTag(endCode, args))
+        case '\r' =>
+        // ignore carriage return
+        case c =>
+          builder.append(c)
+      }
+      i += 1
+    }
+    builder.toString()
+  }
+}
--- a/ui/app/activator/HomePageActor.scala
+++ b/ui/app/activator/HomePageActor.scala
@@ -67,7 +67,8 @@ object HomePageActor {
     def apply(info: String): JsValue =
       JsObject(Seq(
         "response" -> JsString("Status"),
-        "info" -> JsString(info)))
+        "info" -> JsString(info),
+        "infoHtml" -> JsString(AnsiToHtml.ansiToHtml(info))))
   }
   case class Respond(json: JsValue)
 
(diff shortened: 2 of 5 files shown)


ui/app/activator/AnsiToHtml.scala
```diff
@@ -0,0 +1,77 @@
...
+            case "31" => "red"
+            case "32" => "green"
+            case "33" => "yellow"
+            case "34" => "blue"
+            case "35" => "magenta"
+            case "36" => "cyan"
+            case "37" => "white"
+            case _ => ""
+          }
+          s"""<span style="color: ${color}">"""
```
corruptmemory: In the case of a non-match we get:

```html
<span style="color: ">
```

I understand that this should be rare, but it sounds like in the case of a non-match we should drop the `span` altogether.

ui/app/activator/AnsiToHtml.scala
```diff
@@ -0,0 +1,77 @@
...
+            case "41" => "red"
+            case "42" => "green"
+            case "43" => "yellow"
+            case "44" => "blue"
+            case "45" => "magenta"
+            case "46" => "cyan"
+            case "47" => "white"
+            case _ => ""
+          }
+          s"""<span style="background-color: ${color}">"""
```
corruptmemory: Same comment here as above.

corruptmemory:
Other than completely minor comments, LGTM.  Happy to merge if @jsuereth is ok with it.

ui/app/activator/AnsiToHtml.scala
havocp: if we drop it entirely we would have an unmatched '</span>' later (not that I'm very rigorous here about absolutely ensuring we pair these, but given sane input they are supposed to be paired). I could drop the invalid `style` though

ui/app/activator/AnsiToHtml.scala
```diff
@@ -0,0 +1,77 @@
...
+          val color = args match {
+            case "40" => "black"
+            case "41" => "red"
+            case "42" => "green"
+            case "43" => "yellow"
+            case "44" => "blue"
+            case "45" => "magenta"
+            case "46" => "cyan"
+            case "47" => "white"
+            case _ => ""
```
jsuereth: 48, then has 3 arguments "r, g, b"

```
48	Reserved for extended set background color	typical supported next arguments are 5;x where x is color index (0..255) or 2;r;g;b where r,g,b are red, green and blue color channels (out of 255)
```
http://en.wikipedia.org/wiki/ANSI_escape_code

ui/app/activator/AnsiToHtml.scala
```diff
@@ -0,0 +1,77 @@
...
+          val color = args match {
+            case "30" => "black"
+            case "31" => "red"
+            case "32" => "green"
+            case "33" => "yellow"
+            case "34" => "blue"
+            case "35" => "magenta"
+            case "36" => "cyan"
+            case "37" => "white"
+            case _ => ""
```
jsuereth: 38 has three arguments, r g b

48, then has 3 arguments "r, g, b"

```
38	Reserved for extended set foreground color	typical supported next arguments are 5;x where x is color index (0..255) or 2;r;g;b where r,g,b are red, green and blue color channels (out of 255)
```
http://en.wikipedia.org/wiki/ANSI_escape_code

Also, 39 is the "standard" color.

ui/app/activator/AnsiToHtml.scala
corruptmemory: Sure, it goes without saying that dropping the `span` would mean dropping it the Right Way(TM) :-)

ui/app/activator/AnsiToHtml.scala
```diff
@@ -0,0 +1,77 @@
...
+    var i = 0
+    while (i < text.length) {
+      text.charAt(i) match {
+        case '<' => builder.append("&lt;")
+        case '>' => builder.append("&gt;")
+        case '"' => builder.append("&quot;")
+        case '\'' => builder.append("&#x27;")
+        case '&' => builder.append("&amp;")
+        case 27 if ((i + 1) < text.length) && (text.charAt(i + 1) == '[') =>
+          // any character in 64-126 ends the sequence (which is mostly
```
jsuereth: This isn't 100% true. YOu may fail to grab certain chracters which are key escapes and typed in somewhere, like left-arrow.

You want to limit the first check to 64-95 and THEN 64-126 after that.  This was a bug in sbt's ANSI escape parsing that lead to weird failures, and I think you've duplicated it.

jsuereth:
A few comments. LGTM.  You can merge after cleanup if you want, but you should try to fix the escape code parsing.

ui/app/activator/AnsiToHtml.scala
havocp: should we fix Wikipedia to explain this or leave the next person to also get it wrong :-)

havocp: I don't understand what you mean by "first check" here though - where do in pass through 96-126 exactly?

jsuereth: Re-read wikipedia. You need to read to read teh non-CSI escape code sequence bit.

jasl:
I think cleaning ANSI escape codes in backend is not a good way.
The server should just pass the STDOUT and let UI do its needs.

havocp:
This is the backend for the frontend though - the real backend is sbt server (which we decided will not clean them).

jasl:
But it made a restriction that its only support for browser-based frontend(consider to implement a native client), and CSS named colours aren't easy to modify, because it's global. Hard coding HTML is be considered as a bad design in most scenarios.

havocp:
oh, I think I see the confusion. A native client would go to sbt server, not to this code.

ui/app/activator/AnsiToHtml.scala
Warry: IMHO, it'd be preferable to add a class name, instead of forcing the color.

jasl: add a class name isn't enough, consider when you refactor css class name, you need to work both with backend and frontend.

jasl:
Maybe you're right, I'm a new guy to learning Scala and researching Activator. and found this issue too.
My idea is let frontend do the job, ANSI code has a strong identity so you can use REGEX to parse it, just keeping about coloured and font style, dropping others. so we just need to work with JS, no need to change both server-side and browser-side.

Warry:
@jasl : Itried, but the browser can receive up to hundreds of messages per seconds, and regexes in Js are very slow. Hence the headaches. In Activator 1.0 we had a lot less logs to take care about so it was quite seamless. It might be ok on chrome with an i7 processor, but it unlikely on a slower browser and smaller processor.

ui/app/activator/AnsiToHtml.scala
Warry: Seems more likely to me that we need to change the colours themselves rather than class names.

jasl:
@Warry : how about buffering STDOUT on backend?

ui/app/activator/AnsiToHtml.scala
jasl: but it may broken somewhere's style

PR #994, opened 2015-04-01
Generalized proxy system

corruptmemory:
[DO NOT MERGE YET]

* Authentication is a cached piece of data like any other
* Any piece of data can depend on any other piece of data.  If
  authentication is needed then user-interaction will proceed in an
  obvious fashion.
* All remote actions report to the UI their status.  Most actions
  support cancel operations
* Moved from snap -> activator package

Do not merge as there are still some problems with the tests that I need to work out (unstable timing-related problems) and the JavaScript API needs to be updated.  This said, please start commenting on code and problems.


ui/app/activator/AppWebSocketActor.scala
```diff
@@ -18,10 +19,24 @@ class AppWebSocketActor(val config: AppConfig) extends WebSocketActor[JsValue] w
   lazy val appDynamicsConfig = AppDynamics.fromConfig(Play.current.configuration.underlying)
   lazy val appDynamicsActor: ActorRef = context.actorOf(monitor.AppDynamicsActor.props(appDynamicsConfig, defaultContext))
   lazy val newRelicActor: ActorRef = context.actorOf(monitor.NewRelicActor.props(NewRelic.fromConfig(Play.current.configuration.underlying), defaultContext))
+  val typesafeComConfig = TypesafeComProxy.fromConfig(Play.current.configuration.underlying)
+  val lookupTimeout = typesafeComConfig.lookupTimeout
+  val loginEndpoint = AuthenticationActor.httpDoAuthenticate(typesafeComConfig.login.url, typesafeComConfig.login.timeout, defaultContext)_
+  val subsciberEndpoint = SubscriptionDataActor.httpGetSubscriptionData(typesafeComConfig.subscriptionData.url, typesafeComConfig.subscriptionData.timeout, defaultContext)_
```
havocp: spelling typo

ui/app/activator/HttpHelper.scala
```diff
@@ -79,12 +79,19 @@ object HttpHelper {
     innerStep(0)
   }
 
+  def identityHolder(holder: WSRequestHolder): WSRequestHolder = holder
+
+  def proxyHolder(holder: WSRequestHolder): WSRequestHolder = (sys.props.get("http.proxyUser"), sys.props.get("http.proxyPassword")) match {
+    case (Some(u), Some(p)) => holder.withAuth(u, p, WSAuthScheme.BASIC) // <- Only viable option?
```
havocp: proxy user/password are different from basic auth, surely? in theory you could have proxy user/pass AND basic auth to the final (non-proxy) server. On master at least there's a method withProxyServer https://github.com/playframework/playframework/blob/master/framework/src/play-ws/src/main/scala/play/api/libs/ws/WS.scala#L379 and the ProxyServer thing can provide proxy user/pass, but surely play should handle this out of the box... it looks like it at least has a config option to do so, though google also turns up some reports that it's broken.

ui/app/activator/typesafeproxy/ActivatorLatestActor.scala
```diff
@@ -0,0 +1,113 @@
...
+    case UIActor.Cancel => cancel()
+    case UIActor.Retry => onRetry()
+  }
+
+  def handleResult(result: Try[ActivatorLatestInfo]): Unit = result match {
+    case x @ Success(_) =>
+      replyTo ! ActivatorInfo.Put(x, version, self)
+      context stop self
+    case Failure(e: ProxyTimeout) =>
+      log.error("Timeout fetching Activator info", e)
```
havocp: this will happen if people are offline or whatever, and the exception won't be useful. I'd probably stick to info level and say "Unable to fetch latest Activator version information" or something more end-user-oriented.

ui/app/activator/typesafeproxy/ActivatorLatestActor.scala
```diff
@@ -0,0 +1,113 @@
...
+      uiActor ! UIActor.RetryableRequests.Failure(e.getMessage, self, true)
```
havocp: should use the parameter keyword name here for the `true`, reading this I have no idea what `true` means (without scrolling around to find this case class). should be `foo=true` for readability.

ui/app/activator/typesafeproxy/ActivatorLatestActor.scala
```diff
@@ -0,0 +1,113 @@
...
+      context.become(onFailure(() => context.become(runRequest())))
+    case x @ Failure(e) =>
+      log.error("Unknown exception fetching Activator info", e)
+      uiActor ! UIActor.RetryableRequests.Failure(e.getMessage, self, false)
```
havocp: same for `false` of course

ui/app/activator/typesafeproxy/SubscriptionDataActor.scala
```diff
@@ -0,0 +1,130 @@
...
+  def handleResult(auth: AuthenticationStates.Authenticated, result: Try[SubscriberData]): Unit = result match {
+    case x @ Success(_) =>
+      replyTo ! SubscriberDetail.Put(x, version, self)
+      context stop self
+    case Failure(e: ProxyInvalidCredentials) =>
+      replyTo ! Authentication.Put(Failure(e), version, self)
+      uiActor ! UIActor.RetryableRequests.Failure(e.getMessage, self, true)
+      context.become(onFailure(() => context.become(getAuthentication())))
+    case Failure(e: ProxyTimeout) =>
+      log.error("Timeout during fetching subscriber data", e)
```
havocp: same point as above, if you're offline or something we don't want to spam stacktraces.

ui/app/activator/typesafeproxy/ActivatorLatestActor.scala
corruptmemory: RE: `foo=true` - good idea.  Could also use types with names as well

RE: timeout and offline - ok.

ui/conf/application.conf
```diff
@@ -67,6 +67,23 @@ console.update-frequency=2000 # milliseconds
 activator {
   default-page-limit = 100
 
+  typesafe-com-proxy {
+    default-timeout = 5 seconds
+    lookup-timeout = 3 seconds
```
havocp: let's make these longer - think australians. I don't think we want to display a failure to the user anyway (and these lookups aren't a result of their action), so it may as well be a long timeout to give best chance of success.

havocp:
I'm wondering "how can this patch be smaller?" and one idea is to avoid knowledge of specific endpoints and JSON parsing on the server side. I guess those are there in case we want to get latest activator or subscription info on the server side, but do we actually ever need it on server side? can we leave all the parsing and error handling to the client and make the server more of just a pass-through?

corruptmemory:
Clearly no, we don't (at the moment) need this server-side.  I guess it's an "in principle" decision.  Then again we could have done this via the fix that @Warry proposed a while back:

```Scala
import play.api.http.HeaderNames._
Ok(_).withHeaders(ACCESS_CONTROL_ALLOW_ORIGIN -> "*")
```

And do everything we want in the client (I think).  The only thing about this approach is it shuts off (or at least makes very awkward) using this data server-side, should we want to.

havocp:
Is this ready to merge or pending anything?

Warry:
What about the ppoup ?

corruptmemory:
On a new PR

Warry:
LGTM then

PR #1000, opened 2015-04-08
Add Partners tab

havocp:


jsuereth:
This LGTM. 

Also, do we know if Play WS  client is using the same kind of proxy-parameters we have to use in sbt?

jsuereth:
cc @corruptmemory Think we should probably figure out one way to proxy typesafe.com though....

corruptmemory:
Firstly, 1000 was to be mine. :-)

Secondly, RE: proxy, yeah, this is another wrinkle.  It's not bad though because it really only applies to the UI and not to any other sbt-related functionality.

ui/app/controllers/api/Proxy.scala
```diff
@@ -0,0 +1,26 @@
...
+import scala.util.control.NonFatal
+import play.api.libs.ws._
+
+object Proxy extends Controller {
+  import play.api.Play.current
+
+  def getTypesafe(path: String) = Action.async { request =>
+    val holder = WS.url(s"https://typesafe.com/$path")
+      .withHeaders("Accept" -> "text/html")
+      .withRequestTimeout(25000)
```
corruptmemory: Configurable parameter?

corruptmemory:
We now have a couple of proxy approaches.  I'be beginning to think we should throw in the towel and do this all client-side.  We can set the CORS headers on each request from the server:

```
Access-Control-Allow-Origin: https://typesafe.com/
Access-Control-Allow-Methods: POST, GET
Access-Control-Allow-Headers: cookie, origin, content-type, accept
```

and then we can just interact with typesafe.com from JavaScript

corruptmemory:
otherwise LGTM

corruptmemory:
GAH! We'd need to add:

```
Access-Control-Allow-Origin: *
```

To Typesafe.com.  Even if we added a CORS pre-flight check on an expected header this is not a good situation unless we don't mind arbitrary JavaScript running on arbitrary browsers making API calls from arbitrary sites.

PR #1001, opened 2015-04-10
Updated workflow for Reactive Platform projects...

corruptmemory:
This is an updated UX flow for Reactive Platform projects

* Needing/creating a `typesafe.properties` file is detected via the `sbt-typesafe-rp` plugin instead of being programmed into activator
* New flow supports getting a subscriber's ID directly via Activator


ui/app/activator/ProjectWatcher.scala
```diff
@@ -91,9 +91,14 @@ class ProjectWatcher(val location: File, val newSourcesSocket: ActorRef, val app
     name.endsWith(".sbt")
   }
 
+  private def isTypesafeProperties(f: File): Boolean = {
+    val name = f.getName
+    name.endsWith("typesafe.properties")
```
havocp: getName gets only the base filename not the full path I think so `==` rather than `endsWith` might be right.

ui/app/assets/widgets/header/header.js
```diff
@@ -17,15 +21,50 @@ define([
...
   tpl
 ){
+  var showBanner = ko.computed(function () {
+    var irpp = sbt.tasks.reactivePlatform.isReactivePlatformProject();
+    var pfe = sbt.tasks.reactivePlatform.propertiesFileExists();
+    var sid = sbt.tasks.reactivePlatform.subscriptionId();
+    return irpp && (!pfe || !sid);
+  });
+
+  function remedy() {
```
havocp: I think `remedy` gets set on the global object, unless require.js somehow blocks that - elsewhere we do `var remedy = function() { }`. I'm not sure it's actually necessary though.

ui/app/views/main.scala.html
```diff
@@ -49,5 +49,15 @@
     </li>
   </script>
 
+  <script id="missingProperties" type="text/html">
+    <p>Opened a <i>Reactive Platform</i> project missing a <span class='properties-filename'>typesafe.properties</span> file. Click <button class="button" data-bind='click: remedy'>here</button> to fix.</p>
```
havocp: how about `this Reactive Platform project needs a typesafe.properties file, click here to add one` ?

("fix" sounds broken, but this is a normal/expected path; "opened" is awkward passive voice).

same for the other message below.

corruptmemory: Well, we could say "You opened ..." ;-)

No objections to the proposed changes.

corruptmemory:
OK, this PR has been significantly updated and is ready for review.

Ping: @havocp and @jsuereth 

I will update the story as well, but modulo checking for "truly" invalid IDs this is about ready to go.

ui/app/assets/widgets/typesafeIdForm/typesafeIdForm.html
```diff
@@ -0,0 +1,23 @@
+<div data-bind="visible: visible" id="typesafe-id-form">
+  <div class="modal-wrapper">
+    <header>
+      <h2>Submit your Typesafe ID</h2>
+    </header>
+    <article>
+      <p>You are opening a Typesafe Reactive Platform project, which requires a Typesafe ID.</p>
+      <p>You can retrieve your ID, or sign up for a free trial, on the <a href='https://typesafe.com/product/typesafe-reactive-platform/id' target='_blank'>subscription ID page</a> page.</p>
```
havocp: `page</a> page`

also I think we usually call it "subscription ID" not Typesafe ID

havocp:
Should this be reviewed and merged yet or no?

corruptmemory:
@jsuereth wants one more `Cancel` button :-).  Just about ready.

corruptmemory:
OK, ready to kick the tires.

PR #1006, opened 2015-04-14
Use sbteclipse 3.0.0

benmccann:
Eclipse 3.0.0 is being used in ./project/project/eclipse.sbt and ./project/eclipse.sbt. We should use it everywhere. Older versions are evil and will wreak havoc on your sbt build.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -30,7 +30,7 @@ object Dependencies {
 
   // versions used by build to set dependencies in JS
   val ideaVersion = "1.6.0"
-  val eclipseVersion = "2.5.0"
+  val eclipseVersion = "3.0.0"
 
   val activatorCommon      = "com.typesafe.activator" % "activator-common" % templateCacheVersion
   val templateCache        = "com.typesafe.activator" % "activator-templates-cache" % templateCacheVersion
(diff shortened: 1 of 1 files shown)


havocp:
my only question here is what it does to "offlineTests" (hopefully nothing)

benmccann:
How do I check that? Run `sbt offlineTests`? Assuming that's right, I tried that and it doesn't work for me because I'm on Java 8. Do you think activator can move to Java 8 like Play and Akka have?

benmccann:
sbteclipse 4.0.0 is almost out. We really need to stop using 2.5.0, which was broken in very evil ways that would cause Eclipse to screw with SBT such that you would have tons of impossible to diagnose compilation problems in SBT

jsuereth:
Hey @benmccann `offlineTests` is the task you run to try things out, have to be on Java 6.   Should we just skip to sbteclispe 4.0.0?

We need to cut a new release soon, so I'd like to figure this out now that we have 1.3.3 out.

benmccann:
I can't run that test since I only have Java 8 installed. Upgrading to sbteclipse 4.0.0 sounds great. It's only on RC2 at the moment, but we should release 4.0.0 final since Play 2.4 depends on it and was just released.  Can you cut the release? I'm afraid I'm way too busy at the moment to debug why I can't publish

corruptmemory:
Closed.  Replaced by #1082

PR #1024, opened 2015-05-22
Updated echo and sbt-echo publication and offline updates

corruptmemory:
* Upgraded to working template cache hash


henrikengstrom:
LGTM

PR #1031, opened 2015-05-27
Added support to retry "not a subscriber" case

corruptmemory:
* "Not a subscriber" is now indicated in the UI
* You can try another account 
* Fixed parsing Typesafe.com response for subscription API (changed upstream)
* Fixed two possible race conditions in JS.

### DIFF
--- a/ui/app/assets/widgets/typesafeIdForm/typesafeIdForm.html
+++ b/ui/app/assets/widgets/typesafeIdForm/typesafeIdForm.html
@@ -8,6 +8,7 @@
       <p>You can retrieve your ID, or sign up for a free trial, on the <a href='https://typesafe.com/product/typesafe-reactive-platform/id' target='_blank'>subscription ID page</a> page.</p>
       <div class="input-wrapper"><div data-bind="css: typesafeIdStateClass"></div><input class='typesafeId-form' type='text' data-bind="textInput: typesafeId" /></div>
       <p class='typesafe-id-error' data-bind='visible: errorVisible, html: typesafeIdError'></p>
+      <p class='typesafe-id-error' data-bind='visible: notASubscriber'>Your account does not have a Typesafe subscription.  You may wish to try another account.</p>
       <!-- ko if: continueVisible -->
       <p class='get-from-typesafe'><button class='from-typesafe-button' data-bind="click: checkId">Continue</button></p>
       <!-- /ko -->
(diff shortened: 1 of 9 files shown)


ui/app/assets/widgets/typesafeIdForm/typesafeIdForm.html
```diff
@@ -8,6 +8,7 @@
       <p>You can retrieve your ID, or sign up for a free trial, on the <a href='https://typesafe.com/product/typesafe-reactive-platform/id' target='_blank'>subscription ID page</a> page.</p>
       <div class="input-wrapper"><div data-bind="css: typesafeIdStateClass"></div><input class='typesafeId-form' type='text' data-bind="textInput: typesafeId" /></div>
       <p class='typesafe-id-error' data-bind='visible: errorVisible, html: typesafeIdError'></p>
+      <p class='typesafe-id-error' data-bind='visible: notASubscriber'>Your account does not have a Typesafe subscription.  You may wish to try another account.</p>
```
henrikengstrom: Suggestion "This account" instead of "Your account". Also has two spaces after the dot.

ui/app/assets/widgets/typesafeIdForm/typesafeIdForm.js
```diff
@@ -17,6 +17,7 @@ define([
   var FormState = (function(){
     var self = {};
     self.dummyRecompute = ko.observable(1);
+    self.notASubscriber = ko.observable(false);
```
henrikengstrom: Without having background info: 
Why not call this ``subscriber`` instead? Also, why is this set to ``false`` as default? Shouldn't it be the other way around, i.e. you're not a subscriber until proven otherwise?

henrikengstrom:
LGTM apart from the questions above.

ui/app/assets/widgets/typesafeIdForm/typesafeIdForm.js
corruptmemory: Simple really: I suck at names?

Actually, because I cut-n-pasted the name from elsewhere.  It could be something like `showNotASubscriberMessage`.  It's only used to control the visibility of a form element.

henrikengstrom: Don't sweat it. Hardest problem in SW :)

PR #1041, opened 2015-06-09
Buncha fixes

corruptmemory:
* Fixes setting the correct Play fork run version
* Removed Echo/Inspect

review by: @jsuereth


corruptmemory:
Just as a note: this PR means moving to Java 8 for running Activator due to the Play plugin requiring Java 8 (since their run-support artifact was compiled with Java 8)

jsuereth:
This LGTM, sorry it took so long to review!

PR #1061, opened 2015-07-30
Use subshell instead of exec for execRunner

bowmantd:
The bash builtin exec will take over the current shell so an alternate solution is to run the command in a subshell which will allow the command to complete and then resume processing the script.

This will fix the tainted cygwin tty since the cleanup commands from the script are never ran as it is.


typesafehub-validator:
Hi @bowmantd,

Thank you for your contribution! We really value the time you've taken to put this together.

Before we proceed with reviewing this pull request, please sign the Typesafe Contributors License Agreement:

[http://www.typesafe.com/contribute/cla](http://www.typesafe.com/contribute/cla)

jsuereth:
@bowmantd Thanks for the fix.  Just curious, does this break teh ability to CTRL+C the script?

bowmantd:
@jsuereth No, Ctrl-C will still function as before.  I did just submit another change to better handle Ctrl-D.  Normally, Ctrl-D send the logout character which forces the shell to close.  The activator script will now trap for Ctrl-D and stop the subshell but continue the parent script to allow for cleanup specific to cygwin.

PR #1068, opened 2015-08-14
Fix `file:` URIs under Windows

corruptmemory:
This PR will write `file:` URIs under Windows as `file:////` instead of `file://`.  This scheme successfully makes use of the offline repo.

### DIFF
--- a/launcher/src/main/scala/activator/RepositoryConfig.scala
+++ b/launcher/src/main/scala/activator/RepositoryConfig.scala
@@ -75,6 +75,8 @@ import java.io.File
 object RepositoryConfig {
   private val repositoriesSectionName = "repositories"
 
+  private val isWindows = sys.props("os.name").toLowerCase.indexOf("win") >= 0
+
   // configure your per-user repos to have the offline
   // repo in activator.home, if possible.
   def configureUserRepositories(): Unit =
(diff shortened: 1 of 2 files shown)


eed3si9n:
> Do not add (and remove) local local resolvers under Windows #1068

To distinguish with `local` repository (`~/.ivy2/local/`), could we call `activator-local` repository as "preloaded local repository", or "activator-local repository"?

Also to be clear, this fix just removes the `repositories` entry, but the preloaded local repo will still be shipped (for now)?

corruptmemory:
Yes, we're referring to the lines in `~/.sbt/repositories` that look like:

```
activator-launcher-local: file://${activator.local.repository-${activator.home-${user.home}/.activator}/repository}, [organization]/[module]/(scala_[scalaVersion]/)(sbt_[sbtVersion]/)[revision]/[type]s/[artifact](-[classifier]).[ext]
activator-local: file://${activator.local.repository-//C:/Users/jim/Desktop/activator-dist-1.3.5/repository}, [organization]/[module]/(scala_[scalaVersion]/)(sbt_[sbtVersion]/)[revision]/[type]s/[artifact](-[classifier]).[ext]
```

The preloaded local repo is still shipped, but not used.

corruptmemory:
Hrm.  Going to experiment with this some more.  I think there may be a better fix.  Don't merge yet.

corruptmemory:
OK, this PR has had a major update.  @jsuereth, @eed3si9n please have a look. /cc @henrikengstrom

launcher/src/main/scala/activator/RepositoryConfig.scala
```diff
@@ -112,30 +119,33 @@ object RepositoryConfig {
       old
   }
 
+  private lazy val fileMarker = if (isWindows) "file:////" else "file://"
```
eed3si9n: Do you have any link to indicate why/how this works? Maybe add a comment for future maintainer?

corruptmemory: Good point.

So 1) users figured out that `file:///` works.  2) I noticed that the Activator launcher generates log lines with `file:////` in them. and 3) there is this: https://github.com/typesafehub/activator/blob/master/dist/src/templates/activator.bat#L209

So, I think that either `file:///` or `file:////` will work as expected under Windows.

TL;DR: I'll add a comment.

PR #1071, opened 2015-08-30
isLocal is a boolean, not a function; fixes #1029

### LINKED ISSUE #1029: Cannot view tutorials in activator ui
With the 1.3.3 release, I can no longer view local activator template tutorials from the activator ui.

timcharper:
Confirmed fix in my local staged version of Activator.


corruptmemory:
Looks good.  Thanks for the fix.

PR #1075, opened 2015-09-03
Updated template hash to latest

corruptmemory:


corruptmemory:
Trivial update, pushing through.

PR #1080, opened 2015-10-06
Added Project pre-processor to Activator

corruptmemory:
Analyzes projects before they are loaded to perform corrective actions
if possible.

### DIFF
--- a/ui/app/activator/App.scala
+++ b/ui/app/activator/App.scala
@@ -14,16 +14,15 @@ import akka.util.Timeout
 final case class AppIdSocketId(appId: String, socketId: UUID)
 
 class App(val id: AppIdSocketId, val config: AppConfig, val system: ActorSystem,
-  val typesafeComActor: ActorRef,
-  val lookupTimeout: Timeout) extends ActorWrapper {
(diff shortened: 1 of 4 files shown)


eed3si9n:
Can this be tested using scripted or some such integration test?

corruptmemory:
In principle yes.  As you can guess, in practice, no, we just didn't do that.

I manually tested using Play 2.4.x-based templates and Play 2.3.x-based templates.  I also verified with RP templates.

For the core logic in https://github.com/typesafehub/activator/blob/wip/project-preprocessor/ui/app/activator/ProjectPreprocessor.scala I don't need to even write a scripted test.  Just tests against the file system would be adequate.

To do a full test requires more than scripted, you'd need Selenium or some other browser-based automation because the file that gets updated is written via Javascript through the Activator server.  The full interaction requires Activator running.

Yeah, just lots of tedious manual testing is what we've done to date.

eed3si9n:
Well, look ok to me then.

henrikengstrom:
Test ran quickly on Windows and if it's working there it can make it anywhere :)
LGTM

corruptmemory:
"Works on Windows" == "Works anywhere" -- made my day. :-)

PR #1081, opened 2015-10-08
Wip/prepare release

corruptmemory:
Updated template hash to latest and updated artifacts references in the build to reflect new template dependencies.

PR #1082, opened 2015-10-08
Updated version of Eclipse plugin to 4.0.0

corruptmemory:
This PR is built on #1081 but only changes the version of the eclipse plugin.


benmccann:
Yes! sbteclipse 2.2.0 is horribly horribly broken and I would not advise anyone to use it

PR #1098, opened 2016-02-26
Fix up license file

hywel-evans:
Tweak the license file to show accurate information.


dist/src/templates/LICENSE.md
```diff
@@ -6,7 +6,7 @@
 
 ## License
 
-Copyright 2013 Typesafe, Inc.
+Copyright 2016 Typesafe, Inc.
```
corruptmemory: Should we change this to Lightbend?

hywel-evans: Yes.  Good catch!

corruptmemory:
Changed to the Lightbend name

PR #1113, opened 2016-04-16
Wip/update launch scripts

corruptmemory:
OK, ready for a real review

### DIFF
--- a/dist/src/templates/activator.bat
+++ b/dist/src/templates/activator.bat
@@ -52,10 +52,25 @@ rem Detect if we were double clicked, although theoretically A user could
 rem manually run cmd /c
 for %%x in (%cmdcmdline%) do if %%~x==/c set DOUBLECLICKED=1
 
+set SBT_HOME=%BIN_DIRECTORY
+
+rem Detect if we were double clicked, although theoretically A user could
+rem manually run cmd /c
+for %%x in (%cmdcmdline%) do if %%~x==/c set DOUBLECLICKED=1
+
+rem FIRST we load the config file of extra options.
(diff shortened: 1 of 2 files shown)


dist/src/templates/activator.bat
```diff
@@ -45,17 +45,32 @@ set ACTIVATOR_HOME=%ACTIVATOR_HOME:~0,-1%
...
+set SBT_HOME=%BIN_DIRECTORY
+
+rem Detect if we were double clicked, although theoretically A user could
+rem manually run cmd /c
+for %%x in (%cmdcmdline%) do if %%~x==/c set DOUBLECLICKED=1
+
+rem FIRST we load the config file of extra options.
+set FN=%SBT_HOME%\..\conf\sbtconfig.txt
+set CFG_OPTS=
+FOR /F "tokens=* eol=# usebackq delims=" %%i IN ("%FN%") DO (
```
henrikengstrom: I guess this has been tested on Windows?

henrikengstrom:
Without having run this on Windows it LGTM. Merge at will.

dist/src/templates/activator.bat
corruptmemory: Yeah, I did test it under Windows.  I get a warning from that line.  Not sure exactly what to do here.  Assuming folks have sbt (and therefore sbt settings) then why are they using Activator?  This stuff is kinda the equivalent of what is needed here: https://github.com/lagom/lagom/issues/32.  The assumption is that there would be instructions provided by the template as to what to put into the `sbtconfig.txt` file.

PR #1131, opened 2016-08-03
Fix ACTIVATOR_OPTS is ignored on Mac OS and Linux

peter-fu:


peter-fu:
I have signed the CLA.

